PR #29, opened 2022-09-27
cpu-o3: Change degree-based FDIP to stream length-based FDIP

huanglingfeng:
Change-Id: Ibfd1a988f90b6fc6cfda81d4c935bbe9c138d010
Fixed translation bug in tlb of prefetch, change degree-based FDIP to stream length-based FDIP


shinezyy:
Can you split this patch into two commits?
1. Fix translation bug in tlb of prefetch
2. Change degree-based FDIP to stream length-based FDIP

PR #161, opened 2024-08-17
arch-riscv: Do not set mstatus.VS at boot

cyyself:
To align with NEMU and XiangShan RTL to prevent difftest errors, we keep the VS=0 on mstatus at boot.

Change-Id: I1cf440074424787bc63f7d68740ddb1b32e2f526


shinezyy:
@tastynoob @xyyy1420  plz update config used for NEMU and update ref so of CI

xyyy1420:
i have check the config for NEMU, no need to update, everything is ok

cyyself:
Fixed #164

PR #163, opened 2024-08-19
Fix some fp insts not modifying mstatus.fs

lqr8:


shinezyy:
Before send patch, please rebase your code on top of current xs-dev

PR #164, opened 2024-08-20
arch-riscv: default disable vs bit, set misa with B-ext, update ref-so

tastynoob:


src/arch/riscv/isa.cc
```diff
@@ -517,20 +517,20 @@ ISA::setMiscReg(int misc_reg, RegVal val)
...
+                //         bits(tc->pcState().as<RiscvISA::PCState>().npc(),
+                //             2, 0) != 0) {
+                //     val |= cur_val & ISA_EXT_C_MASK;
+                // }
+
+                // if ((val & ISA_EXT_H_MASK) != 0) {
+                //     // Do not allow to enable RVH because not implemented yet
+                //     val &= ~ISA_EXT_H_MASK;
+                // }
+                // setMiscRegNoEffect(misc_reg, val);
```
shinezyy: delete useless code

shinezyy (changes requested):

shinezyy (approved):

PR #165, opened 2024-08-26
Fix some bugs to run splash3 benchmarks

lqr8:


configs/common/FSConfig.py
```diff
@@ -678,11 +678,11 @@ def makeBareMetalXiangshanSystem(mem_mode, mdesc=None, cmdline=None, np=1, ruby=
     self.uartlite  = UartLite()
     self.uartlite.pio = self.iobus.mem_side_ports
 
-    self.lint = Lint()
+    self.lint = Clint()
     self.lint.pio = self.iobus.mem_side_ports
     self.lint.pio_addr = 0x38000000
     self.lint.num_threads = np
-    self.lint.int_enable = 1
+    # self.lint.int_enable = 1
```
shinezyy: delete useless code

src/arch/riscv/isa.cc
```diff
@@ -138,6 +138,11 @@ namespace RiscvISA
     [MISCREG_HPMEVENT29]    = "HPMEVENT29",
     [MISCREG_HPMEVENT30]    = "HPMEVENT30",
     [MISCREG_HPMEVENT31]    = "HPMEVENT31",
+
+    [MISCREG_MTINST]        = "MTINST",
+
+    [MISCREG_MTVAL2]        = "MTVAL2",
+
```
shinezyy, line 144: plz remove extra spaces

src/arch/riscv/isa/decoder.isa
```diff
@@ -1660,29 +1660,41 @@ decode QUADRANT default Unknown::unknown() {
                         auto sign = bits(unboxF32(Fs2_bits), 31);
                         Fd_bits = boxF32(insertBits(unboxF32(Fs1_bits), 31,
                                                     sign));
+                        status.fs = 3;
```
shinezyy, line 1663: @tastynoob Is it possible to update status register in templates of floating point instructions instead of modify every instructions?

src/arch/riscv/regs/misc.hh
```diff
@@ -132,6 +132,10 @@ enum MiscRegIndex
     MISCREG_HPMEVENT29,
     MISCREG_HPMEVENT30,
     MISCREG_HPMEVENT31,
+    MISCREG_MTINST,
+
```
shinezyy, line 135: plz remove spaces

src/cpu/o3/fetch.cc
```diff
@@ -1173,6 +1163,21 @@ Fetch::tick()
...
+            issuePipelinedIfetch[0] = false;
+        }
+
+        if (fromCommit->commitInfo[0].clearInterrupt) {
+            DPRINTF(Fetch, "Clear interrupt pending.\n");
+            interruptPending = false;
+            issuePipelinedIfetch[0] = true;
+        }
+    }
+
```
shinezyy, line 1178: Is it necessary to move this code block here from above ?

shinezyy (changes requested):

src/arch/riscv/isa/decoder.isa
tastynoob: yes，see format FPROp

tastynoob: But it seems that this is not recommended, because FPROp is used by many instructions

src/cpu/o3/fetch.cc
lqr8: Yes, `fetch()` above modifies `issuePipelinedIfetch[]`. This code is still buggy so I fixed it again below

src/cpu/base.cc
```diff
@@ -909,19 +909,19 @@ BaseCPU::diffWithNEMU(ThreadID tid, InstSeqNum seq)
     auto gem5_pc = diffInfo.pc->instAddr();
     diffAllStates->gem5RegFile.pc = gem5_pc;
     auto nemu_pc = diffAllStates->diff.nemu_commit_inst_pc;
-    DPRINTF(Diff, "NEMU PC: %#10lx, GEM5 PC: %#10lx, inst: %s\n", nemu_pc,
-            gem5_pc,
+    diffMsg += csprintf("In CPU%d: NEMU PC: %#10lx, GEM5 PC: %#10lx, inst: %s\n", cpuId(),
```
tastynoob: It is not recommended to do this, because diffwithNemu was very frequently called, string operations will be very time consuming

shinezyy: I think it is fine if performance overhead is less than 5%.

diffMsg style brings flexibility

tastynoob (approved):

tastynoob (approved):

shinezyy (approved):

PR #179, opened 2024-10-11
misc: add difftest-ci

jueshiwenli:
make cpu commit the wrong result to check whether difftest work or not

Change-Id: I8e88afee24da5f0d7617e6fa73ae99f9da94e674


shinezyy (approved):
LGTM

PR #189, opened 2024-10-16
Multiple bug fixes to run parsec benchmark

lqr8:


shinezyy:
Clean solution for TLB priv level !

shinezyy (approved):

src/cpu/o3/fetch.cc
```diff
@@ -835,7 +835,7 @@ Fetch::finishTranslation(const Fault &fault, const RequestPtr &mem_req)
         // If we have, just wait around for commit to squash something and put
         // us on the right track
         if (!cpu->system->isMemAddr(mem_req->getPaddr())) {
-            warn("Address %#x is outside of physical memory, stopping fetch, %lu\n",
+            DPRINTF(Fetch, "Address %#x is outside of physical memory, stopping fetch, %lu\n",
```
tastynoob, line 838: Use "warn" can help us find the fetch stuck easier

shinezyy: > Use "warn" can help us find the fetch stuck easier

But if on the speculative path, might it be too verbose?

@notlqr Did you observe a too verbose print here?

lqr8: Some benchmarks produce too much warns here

PR #192, opened 2024-10-21
Fix some minor issues

eastonman:
should be rebase merged.

### DIFF
--- a/src/cpu/o3/issue_queue.cc
+++ b/src/cpu/o3/issue_queue.cc
@@ -186,7 +186,7 @@ IssueQue::checkScoreboard(const DynInstPtr& inst)
                 panic("dst is not load");
             }
             scheduler->loadCancel(dst_inst);
-            DPRINTF(Schedule, "[sn %lu] %s can't get data from bypassNetwork, dst inst: %s\n",
+            DPRINTF(Schedule, "[sn:%llu] %s can't get data from bypassNetwork, dst inst: %s\n",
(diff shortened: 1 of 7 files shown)


src/cpu/o3/issue_queue.cc
```diff
@@ -186,7 +186,7 @@ IssueQue::checkScoreboard(const DynInstPtr& inst)
                 panic("dst is not load");
             }
             scheduler->loadCancel(dst_inst);
-            DPRINTF(Schedule, "[sn %lu] %s can't get data from bypassNetwork, dst inst: %s\n",
+            DPRINTF(Schedule, "[sn:%llu] %s can't get data from bypassNetwork, dst inst: %s\n",
```
tastynoob, line 189: I actually did this on purpose because this can help us find scheduler‘s log from a lot irrelevant information easier and faster.
It is very useful in scheduler debugging.

tastynoob (approved):

src/cpu/o3/issue_queue.cc
eastonman: I don't quite understand, can this be achieved by using only the `Schedule` debug flag?
Recently I have experienced some inconvenience tracking instructions through the scheduler when debugging.

tastynoob: ok

shinezyy (approved):

PR #198, opened 2024-10-31
mem: limit dcache tag read port for load/pf

jueshiwenli:
Change-Id: Ife384e9a9ce59a3916ca06b932d228400a525515


shinezyy (approved):
Changes make sense. I approve this patch as soon as it passes CI

Issue #188, opened 2024-10-16
Value prediction development roadmap


zybzzz:
This is the roadmap for Value prediction development, refer to #181 for the general roadmap development of xs-gem5.

# Value prediction roadmap

## Benchmarks
1. general: SPEC06
2. specific: other domain-specific applications.

## Evaluation Metrics
1. SPEC06 weighted score and IPC speedup
2. Value predictor accuracy rate
3. Value predictor coverage rate

## Development Purpose

The current expectation is to integrate the value predictor **without massive changes** to the microarchitecture. The primary goal of the current development of the value predictor is to integrate the **stride-base value predictor** for pre-RTL architecture exploration for kunminghu v3.

### value predictors
- Stride based
  - [ ] Enhanced Stride ( ES,  [from 1st CVP](https://microarch.org/cvp1/papers/Seznec.pdf) )
- Context based
  - [ ] VTAGE ( [from HPCA2014](https://doi.org/10.1109/HPCA.2014.6835952) )
- Hybrid

### Prediction error strategy
- pipeline flushing.(**current implementation**)
- Add replay-buffer, re-issue from replay-buffer.

## Develop branchs
1. **valuepred-base** as a **main** branch of value prediction development.
2. **valuepred-(feature)** develops different predictors as a **downstream branch** of **valuepred-base**.

**Does valuepred-base need to be synchronized with the upstream branch xs-dev? Any ideas?**

## Future
1. co-design with prefetch
5. co-degisn with memory dependency prediction unit

zybzzz:
@shinezyy @jensen-yan 

Preliminary performance data for the value predictor is given, currently only **overall** scores are given, detailed data has not yet been analyzed. A 5KB stride predictor is given first, as there are still bugs associated with the current predictor design. In comparison, the performance **yields a 42% miss**, which is **very bad**. The exact reason for this is yet to be analyzed.

> ================ Overall baseline=================
> Estimated overall score per GHz: 16.200665687457658
> Estimated overall score @ 3.0GHz: 48.60199706237297

> ================ Overall with 5KB EStride VP=================
> Estimated overall score per GHz: 9.19388382981961
> Estimated overall score @ 3.0GHz: 27.581651489458828

zybzzz:
The ideal model has passed all checkpoint tests on the server. Applying the ideal model to a 100% accurate stride predictor (for ALU operations) and a 100% accurate load value predictor (for load instructions) has yielded the following results:

> baseline:
[...]

PR #201, opened 2024-11-07
basic implement for estride value predictor

zybzzz:
related roadmap: #188 

This PR is solely intended to track code related to value prediction and does not currently need to be merged into xs-dev.

**Implemented features in the current code:**

1. Modified the o3 CPU architecture to support value prediction.
2. Designed a basic interface for the value predictor and implemented a simple Estride value predictor.
3. Configured related parameters for Estride, including the number of bits saved for stride and the size of the inflight window, allowing for basic design space exploration.

**Known issues:**
1. The current Estride update strategy is still very basic. **A new update strategy based on instruction execution time is needed**, aligning with the original Estride design.
2. when entry size  <= 16 will raising fault.

spec benchmark scoring is still ongoing.

### DIFF
--- a/src/cpu/o3/cpu.cc
+++ b/src/cpu/o3/cpu.cc
@@ -1251,6 +1251,10 @@ CPU::instDone(ThreadID tid, const DynInstPtr &inst)
         thread[tid]->threadStats.numInsts++;
         cpuStats.committedInsts[tid]++;
 
+        if (thread[tid]->numInst % 10000 == 0) {
+            fprintf(stderr, "commit %lu insts\n", thread[tid]->numInst);
+        }
+
         if (this->nextDumpInstCount
                 && totalInsts() == this->nextDumpInstCount) {
             fprintf(stderr, "Will trigger stat dump and reset\n");
--- a/src/cpu/o3/cpu.cc
+++ b/src/cpu/o3/cpu.cc
@@ -60,6 +60,7 @@
 #include "debug/Drain.hh"
 #include "debug/O3CPU.hh"
 #include "debug/Quiesce.hh"
+#include "debug/VPCOMMON.hh"
 #include "debug/ValueCommit.hh"
 #include "enums/MemoryMode.hh"
 #include "sim/async.hh"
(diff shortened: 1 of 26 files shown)


src/cpu/o3/cpu.cc
```diff
@@ -1251,6 +1251,10 @@ CPU::instDone(ThreadID tid, const DynInstPtr &inst)
         thread[tid]->threadStats.numInsts++;
         cpuStats.committedInsts[tid]++;
 
+        if (thread[tid]->numInst % 10000 == 0) {
+            fprintf(stderr, "commit %lu insts\n", thread[tid]->numInst);
```
shinezyy: We always prefer DPRINTF rather than fprintf in GEM5

src/cpu/valuepred/enhanced_stride.cc
```diff
@@ -0,0 +1,553 @@
...
+
+VPResult
+EStride::doPredict(ESPredMetaData *esPredMetaData, int inflights)
+{
+    std::string indexMarker, tagMarker;
+    std::vector<unsigned> indexEachWays(ways);
+    for (size_t i = 0; i < indexEachWays.size(); i++) {
+        indexEachWays[i] = pcHashToWayIndex(esPredMetaData->pc, i);
+        indexMarker += std::to_string(indexEachWays[i]);
+        indexMarker += " ";
```
shinezyy: I'm not sure how such string operation affect performance. If performance is degraded, I suggest to put strong operations into if block like  `if (debug::xxx) {indexMarker = `

shinezyy (changes requested):

shinezyy:
- Did you provide some method to turn off the value predictor?
- If possible, you can provide a preliminary result.

zybzzz:
> * Did you provide some method to turn off the value predictor?
> * If possible, you can provide a preliminary result.

I don't have a way to turn off the value predictor yet, the only way to turn off the value predictor right now is to checkout to xs-dev and turn off the value predictor by changing the branch to cancel the value predictor related changes.

I think eventually I'll implement turning off the value predictor in code.

src/cpu/o3/cpu.cc
zybzzz: I've fixed the problem with this part of the code and I'll start running the scores after I get **issue1** finished.

src/cpu/valuepred/enhanced_stride.cc
zybzzz: I've fixed the problem with this part of the code and I'll start running the scores after I get **issue1** finished.

PR #203, opened 2024-11-13
mem-cache: add l2l3 tag readlimit, add store stall when writing cache

tastynoob:


tastynoob:
image
performance compare

tastynoob:
image

tastynoob:
And optimize the behavior of read tag fail retry, no need to modify lsqunit

src/mem/cache/base.hh
```diff
@@ -1054,6 +1061,20 @@ class BaseCache : public ClockedObject, CacheAccessor
...
+        return (addr >> blkAddrShift) & 0b11;
+    }
+
+    void writeBlock(Addr addr, Tick when) {
+        lastWriteFinishTick[getBank(addr)] = when + cyclesToTicks(dataLatency + Cycles(1));
+    }
+
+    bool isWriteBlock(Addr addr) {
+        return lastWriteFinishTick[getBank(addr)] > curTick();
+    }
```
happy-lx, line 1076: RTL's SRAM only blocks all read requests during the single writing cycle. It seems that `isWriteBlock` blocks all requests from the time the write request is received to the completion of writing?

PR #205, opened 2024-11-14
cpu-o3: reduce int div latency to align kmh

tastynoob:
Align with kmh-v2 div latency


tastynoob:
image
The Data of div test

shinezyy (approved):

jueshiwenli (approved):

PR #206, opened 2024-11-18
cpu-o3: add perfCCT frame, add inst lifetime counter

tastynoob:


src/cpu/o3/commit.cc
```diff
@@ -1676,14 +1678,6 @@ Commit::commitHead(const DynInstPtr &head_inst, unsigned inst_num)
     // Finally clear the head ROB entry.
     rob->retireHead(tid);
 
-#if TRACING_ON
-    if (debug::O3PipeView) {
-        head_inst->commitTick = curTick() - head_inst->fetchTick;
-        DPRINTF(O3PipeView, "Record commit for inst sn:%lu, commitTick=%lu\n",
```
jueshiwenli, line 1682: It is better to keep o3pipview.

src/cpu/o3/perfCCT.hh
```diff
@@ -0,0 +1,73 @@
...
+        AtRename,
+        AtDispQue,
+        AtIssueQue,
+        AtIssueArb,
+        AtIssueReadReg,
+        AtFU,
+        AtBypassVal,
+        AtWriteVal,
+        AtCommit,
+        InvalidPos
```
shinezyy: Can we define this in Python's enumeration and use it when create SLQ table to avoid redundant code?

shinezyy (changes requested):

jueshiwenli (approved):

jueshiwenli (approved):

shinezyy (approved):
Redudant codes are now removed, excellent!

dzwduan:
Why not consider using https://github.com/shioyadan/Konata for pipeline visualization ?

Issue #207, opened 2024-11-18
Add docker support to quickly create a runtime environment on the server


zybzzz:
I'm having a lot of trouble running xs-gem5 on the server. When developing locally, we were able to quickly build the environment we needed for runtime with a package manager such as **apt**. But when we need to run on the server, we often can't get root access to quickly set up the runtime environment, which makes it very difficult to run xs-gem5 in parallel on the server to run scores.

I would like to try to add **docker** support for xs-gem5 in the ```util/xs_scripts``` directory to make it possible to quickly build environments for running scores on a server with docker. I'm sure this will also help for newer users.

PR #208, opened 2024-11-19
add docker support for xs-gem5

zybzzz:
ref issue: #207 

It works fine on my server.

@shinezyy @jensen-yan

### DIFF
--- a/util/xs_scripts/docker/prepare-env.sh
+++ b/util/xs_scripts/docker/prepare-env.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+CHECKPOINTS_DIR="checkpoints"
+RESULTS_DIR="results"
+GCPT_RESTORER="/home/zybzzz/proj/openxiangshan/tools/restorer/gcpt.fromlibcptalpha"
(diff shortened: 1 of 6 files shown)


util/xs_scripts/docker/prepare-env.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+CHECKPOINTS_DIR="checkpoints"
+RESULTS_DIR="results"
+GCPT_RESTORER="/home/zybzzz/proj/openxiangshan/tools/restorer/gcpt.fromlibcptalpha"
```
jensen-yan, line 5: I think these two lines should not appear here,  this way it would only run on your machine, right?

shinezyy:
Not tested, because we don't have docker permission on our servers.
But Code look good to me

shinezyy (approved):

jensen-yan (approved):

util/xs_scripts/docker/prepare-env.sh
zybzzz: > I think these two lines should not appear here, this way it would only run on your machine, right?

Sorry, I didn't see it earlier. Yes, but I still keep it here just to provide an example. @jensen-yan

PR #212, opened 2024-11-26
cpu-o3: add `--disable-sc` param and fix a Segmentation Fault

Lawrence-ID:
* add `--disable-sc` command line parameter to disable SC.
* Fix segmentation fault when using `--enable-bp-db tage` with `enableSC=true`

### DIFF
--- a/configs/common/Options.py
+++ b/configs/common/Options.py
@@ -269,6 +269,8 @@ def addCommonOptions(parser, configure_xiangshan=False):
                         help="enable bp database for specified subdatabase, "
                         "basic branch trace is enabled by default even without specifying, "
                         "available subdatabase: basic, tage, ras, loop")
+    parser.add_argument("--disable-sc", default=False, action="store_true",
+                        help="enable sc (only for ftb branch predictor)")
     parser.add_argument("--enable-loop-buffer", default=False, action="store_true",
                         help="enable loop buffer (only for ftb branch predictor)")
     parser.add_argument("--enable-loop-predictor", default=False, action="store_true",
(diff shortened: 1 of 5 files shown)


configs/example/fs.py
```diff
@@ -313,6 +313,7 @@ def build_test_system(np):
                                                     enableLoopPredictor=args.enable_loop_predictor,
                                                     enableJumpAheadPredictor=args.enable_jump_ahead_predictor
                                                     )
+                    test_sys.cpu[i].branchPred.tage.enableSC = (not args.disable_sc)
```
shinezyy: Why parentheses around "not args...."

configs/example/xiangshan.py
```diff
@@ -89,6 +89,7 @@ def build_test_system(np, args):
                                             enableLoopPredictor=args.enable_loop_predictor,
                                             enableJumpAheadPredictor=args.enable_jump_ahead_predictor
                                             )
+            test_sys.cpu[i].branchPred.tage.enableSC = (not args.disable_sc)
```
shinezyy: Why parentheses around "not args...."?

shinezyy (changes requested):

configs/common/Options.py
```diff
@@ -269,6 +269,8 @@ def addCommonOptions(parser, configure_xiangshan=False):
                         help="enable bp database for specified subdatabase, "
                         "basic branch trace is enabled by default even without specifying, "
                         "available subdatabase: basic, tage, ras, loop")
+    parser.add_argument("--disable-sc", default=False, action="store_true",
+                        help="enable sc (only for ftb branch predictor)")
```
jensen-yan: maybe we should use --enable-sc, not --disable-sc, help= "enable sc(for tage predictor)", this code writing style is more unified.

jensen-yan (approved):

shinezyy (approved):

PR #218, opened 2024-12-02
mem: align cache latency use maprobe

tastynoob:
use maprobe benchmark to align cache latency with rtl


jensen-yan:
According to the CI performance data, bzip2 gets lower score( 7.822  vs 7.975), others get  higher score. Int avg(15.73 vs 15.40).
Cache latency is aligned to RTL, however bandwidth is not aligned. 
Maybe we should wait for bandwidth aligning.

jueshiwenli (commented):
maprobe may has Some issues are being analyzed, especially in l3, i am checking the benchmark

jueshiwenli (approved):

PR #221, opened 2024-12-03
util: add perfcct color

jensen-yan:
Change-Id: I645c22a0d87d71556eb085a0e4f9d650de108ada


jensen-yan:
No need to run CI

jensen-yan:
Since it may output to file, adding color may cause other problems, only output to terminal is acceptable.

PR #224, opened 2024-12-10
cpu-o3: reg arb smaller value hignher priority

tastynoob:


tastynoob:
image

tastynoob:
before: bigger value get higher priority, bru will reject alu
after: smaller value get higher priority, alu will reject bru

jensen-yan (approved):

PR #225, opened 2024-12-10
configs: pick prefetcher config from cache config

### LINKED ISSUE #196: Feature request: Unify the configuration of prefetchers
Currently, prefetcher configuration is scattered across multiple files. For example, when we use `configs/example/kmh.py`, both `configs/common/CacheConfig.py` and `configs/example/kmh.py` will configure part of prefetchers, which is confusing.

We should put all prefetcher configuration code into one script.

tastynoob:


shinezyy (approved):
lgtm

PR #226, opened 2024-12-11
cpu-o3: refactor KunminghuV3 ideal params configuration

Lawrence-ID:
This commit is NOT expected to impact performance.
* Added a new runtime parameter '--ideal-kmhv3'. Parameters reset in the `setKmhV3IdealParams` function now take the highest priority, overriding command-line arguments
* Parameterized `baseTableSize` and `enableSC` in the FTBTAGE class for easier customization
* Unified `numBr` into the DecoupledBPUWithFTB class in BranchPredictor.py for better consistency


eastonman (approved):
LGTM

PR #232, opened 2024-12-13
cpu-o3: branchPred: add ftq end reason stats

jensen-yan:
To understand why  32byte fetchBlock do not have 8 insts, because of size_limit, earlier taken or loop_end.
If size limits, we could enlarge fetchBlock size.


jensen-yan:
No need to run CI.

tastynoob (approved):

PR #233, opened 2024-12-13
Optimize CDP Prefetcher

happy-lx:
+ add degree in cdp (This PR set depth as 1 & degree as 3)
+ restructure the vpnTable
+ add coalescing mechanism in vpnTable
+ add dynamic control & [coordinate throttling](https://people.inf.ethz.ch/omutlu/pub/bandwidth_lds_hpca09.pdf) in cdp
+ add filterTable in cdp to filter out as much not useful prefetch reqs as possible

It is worth mentioning that **the dynamic control mechanism does not necessarily fully adapt to the current architecture** , and **turning on the dynamic control mechanism may cause other changes that are not very relevant to indirectly affect the working status of the dynamic control mechanism.** As a result, when observing performance changes of other changes, the variables are not unique.(may be affected by the dynamic control mechanism) 

Therefore, **if there are some strange performance changes happens, you can consider turning off the dynamic control mechanism and then test again**, by setting `enable_coordinate` to False.


src/mem/cache/prefetch/l2_composite_with_worker.cc
```diff
@@ -26,6 +27,31 @@ L2CompositeWithWorkerPrefetcher::L2CompositeWithWorkerPrefetcher(const L2Composi
...
+    if (pfSource == PrefetchSourceType::CDP) {
+        cdp->recordUnusedPrefetch(paddr);
+    }
+}
+
+void
+L2CompositeWithWorkerPrefetcher::addToQueue(std::list<DeferredPacket> &queue, DeferredPacket &dpp)
+{
+    if (&queue == &pfq) {
+        // DPRINTF(CDPFilter, "addToQueue source: %#llx, vaddr: %#llx, paddr: %#llx\n",
```
jueshiwenli: Remove unnecessary comments.

src/mem/cache/prefetch/cdp.hh
```diff
@@ -71,79 +72,329 @@ namespace prefetch
...
+            uint64_t _rst_period, Entry const &init_val = Entry())
+                : table(assoc, num_entries, idx_policy, rpl_policy, init_val),
+                  _resetPeriod(_rst_period),
+                  _resetCounter(0),
+                  _subEntryNum(sub_entries)
+            {
+                assert(_subEntryNum % 2 == 0 && _subEntryNum < 512);
+                _subEntryBits = ceil(log2(_subEntryNum));
+            }
+            void add(int vpn2, int vpn1, bool pf_hit_cdp) {
```
jueshiwenli: Please use the same code format.(235 257)
(1)void aaa() {
}
(2)void aaa()
{
}

src/mem/cache/prefetch/l2_composite_with_worker.cc
happy-lx: orz

src/mem/cache/prefetch/cdp.hh
happy-lx: Thanks I will fix them. orz

src/mem/cache/prefetch/cdp.cc
```diff
@@ -54,9 +55,19 @@ namespace prefetch
 {
 CDP::CDP(const CDPParams &p)
     : Queued(p),
-      depth_threshold(3),
+      enableCoordinate(p.enable_coordinate),
+      depth_threshold(1),
```
jueshiwenli, line 59: Maybe parameterization would be better, but that's fine

src/mem/cache/prefetch/cdp.cc
```diff
@@ -263,13 +301,17 @@ CDP::notifyWithData(const PacketPtr &pkt, bool is_l1_use, std::vector<AddrPriori
...
                 }
-                vpnTable.update(vpn2, vpn1, enable_thro);
+                vpnTable.update(vpn2, vpn1, enable_thro, isLowConfidence());
                 sendPFWithFilter(blockAddress(test_addr2), addresses, 29 + next_depth, PrefetchSourceType::CDP,
                                  next_depth);
-                if (trueAccuracy > 0.05) {
-                    vpnTable.update(vpn2, vpn1, enable_thro);
-                    sendPFWithFilter(blockAddress(test_addr2) + 0x40, addresses, 1, PrefetchSourceType::CDP,
+                for (int i = 1; i < degree; i++) {
+                    if (trueAccuracy > 0.05) {
```
jueshiwenli, line 308: Maybe parameterization would be better, but that's fine

jueshiwenli (approved):

PR #240, opened 2024-12-17
cpu-o3: ideal: add ideal scheduler, close bank conflict

jensen-yan:


seanwang2046 (approved):
1.set to unified iq for ld & st
2. set to no L1D data bank conflict
3. set to BR num from 4 ->6, FB width from 32B->64B

PR #251, opened 2024-12-30
arch-riscv: fix hpmCounter logic

happy-lx:
0277c32e33b5ff3474a74f488b451ea

- when privilege mode is U, corresponding hpm counter can be read only when corresponding bit of scounteren and mcounteren are both set.
- make scounteren writeable

Change-Id: Ia8332471232db9a2b2ce732ea74dae6497428eef

### DIFF
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -47,6 +47,7 @@
 #include "base/compiler.hh"
 #include "base/logging.hh"
 #include "base/trace.hh"
+#include "base/types.hh"
 #include "cpu/base.hh"
 #include "debug/Checkpoint.hh"
 #include "debug/FloatRegs.hh"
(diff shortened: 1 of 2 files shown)


tastynoob (approved):

eastonman:
I have met a very similar problem recently.
MCOUNTEREN is writable and NEMU `riscv64-gem5-ref_defconfig` sets Zihpm and Zictr off, resulting in difftest error.

After allowing {M,S,H}COUNTEREN to be writable and turning Zihpm Zictr on, no difftest error happens. However, some checkpoints print opensbi trap messages which indicates more problems. 

Do you have any idea about this?

happy-lx:
> I have met a very similar problem recently. MCOUNTEREN is writable and NEMU `riscv64-gem5-ref_defconfig` sets Zihpm and Zictr off, resulting in difftest error. After allowing {M,S,H}COUNTEREN to be writable and turning Zihpm Zictr on, no difftest error happens.

image

yes, I have met very similar things before, I think these options should be turned on in NEMU

> However, some checkpoints print opensbi trap messages which indicates more problems.
> 
> Do you have any idea about this?

984bf07b27bbc95edfe384b577bdbb4

I also encountered an opensbi trap problem. Later I found out that it was because reading the `time` csr on gem5 would always trigger illegal Instruction fault. I’m not sure if you are in a similar situation.

https://github.com/OpenXiangShan/GEM5/blob/d46d2c5c24d82fd78856a3d65eca81b7c30ca753/src/arch/riscv/isa/formats/standard.isa#L365-L371

jueshiwenli (approved):

eastonman:
@happy-lx Yes, my opensbi trap problem is the same as yours.

src/arch/riscv/isa.cc
```diff
@@ -575,10 +576,6 @@ ISA::setMiscReg(int misc_reg, RegVal val)
         auto hcounter = readMiscRegNoEffect(MISCREG_HCOUNTEREN);
         RegVal write_val = ((hcounter & ~(NEMU_COUNTER_MASK)) | (val & NEMU_COUNTER_MASK));
         setMiscRegNoEffect(MISCREG_HCOUNTEREN, write_val);
-    } else if (misc_reg == MISCREG_SCOUNTEREN) {
-        auto scounter = readMiscRegNoEffect(MISCREG_SCOUNTEREN);
-        RegVal write_val = ((scounter & ~(NEMU_COUNTER_MASK)) | (val & NEMU_COUNTER_MASK));
```
eastonman, line 580: I suggest not deleting this logic, but changing the `NEMU_COUNTER_MASK` definition to `0xffffffffULL` ?
Also, I think {M,S,H}COUNTEREN can be moved into the switch-case below because its new value does not involves cpu.v or other information.

happy-lx: Yes, I think the suggestions are very reasonable.

eastonman (approved):
Ok, the code LGTM

PR #252, opened 2025-01-01
arch-riscv: cherry-pick zcb support from upstream

eastonman:
I also removed rv32 related code, since we do not have those APIs.

### DIFF
--- a/src/arch/riscv/isa/bitfields.isa
+++ b/src/arch/riscv/isa/bitfields.isa
@@ -95,7 +95,9 @@ def bitfield RL <25>;
 
 // Compressed
 def bitfield COPCODE <15:13>;
+def bitfield CFUNCT6LOW3 <12:10>;
 def bitfield CFUNCT1 <12>;
+def bitfield CFUNCT1BIT6 <6>;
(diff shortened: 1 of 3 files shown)


tastynoob:
Has it been tested on xs-gem5?

eastonman:
It works with new zcb checkpoint and NEMU with no obvious problem. Here's a comparison between `DeterLoad-gcc1420-zcb-zicond-mimalloc` and the checkpoint we used to use.

Screenshot 2025-01-02 at 15 09 30

eastonman:
Total ckpt amount is 610, and 609 passed difftest with NEMU. The one failed is related to some `ssractch` issue which I believe has nothing to do with Zcb.

tastynoob (approved):

tastynoob:
Looks good

seanwang2046:
significant performance boost got from the support of zicond-mimalloc support. very impressive finding!
we need to verify the performance gain on KMHV2 RTL design, as well as on  KMHV3 GEM5 model; if it works, we need to change the benchmark checkpoint to GCC14 version, this will significantly help to achieve the goal of KMHV3. 
excellent work!

PR #256, opened 2025-01-04
util: add --debug-cycle to help debug print

jensen-yan:
it will enable print cycle in DPRINTF,
print tick is difficult to debug, print cycle helps, DPRINTF print like:
  31635: system.cpu.branchPred: [c: 95] DecoupledBPUWithFTB::tick()
c: 95 means cycle, 95 = 31635 / 333, 333 is ticks for 1 cycle in 3GHz, use cpu->clockPeriod() is greater, but not all class have cpu member.

Change-Id: I0ff1f3e3290842e6ba110393705c003616103924

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -152,6 +152,8 @@ AddOption('--pgo-use', action='store', default=None,
           help='Use pgo profiling results')
 AddOption('--unit-test', action='store_true',
           help='Enable unit test build')
+AddOption('--debug-cycle', action='store_true',
+          help='Enable print cycle in DPRINTF')
 AddOption('--rvv-impl', action='store', default='base',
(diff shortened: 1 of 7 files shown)


happy-lx (dismissed):

shinezyy (changes requested):
The reason that original dprintf printing ticks instead of cycles is that not all clockedobjects are in the same clock domain.

But it is much more convenient for CPU guys to use a CPU core clock (this is the reason of this patch).

To avoid magic 333, I suggest to add a new member function  "cpuCycle()" to "clockedObject".

tastynoob (approved):

boran-xu (approved):

PR #257, opened 2025-01-06
cpu-o3: add intel topdown

tastynoob:


src/cpu/o3/commit.cc
```diff
@@ -1164,8 +1165,15 @@ Commit::commitInsts()
             if (commit_success) {
                 cpu->perfCCT->updateInstPos(head_inst->seqNum, PerfRecord::AtCommit);
                 cpu->perfCCT->commitMeta(head_inst->seqNum);
-                lastCommitCycle = cpu->curCycle();
                 head_inst->printDisassemblyAndResult(cpu->name());
+                if (ismispred) {
```
seanwang2046, line 1189: why let " if(ismisped)..." executed before setting the ismispred =true?

src/cpu/o3/fetch.cc
```diff
@@ -245,7 +245,23 @@ Fetch::FetchStatGroup::FetchStatGroup(CPU *cpu, Fetch *fetch)
     ADD_STAT(decodeStallRate, statistics::units::Rate<
                     statistics::units::Count, statistics::units::Cycle>::get(),
              "Number of decode stalls per cycle",
-             decodeStalls / cpu->baseStats.numCycles)
+             decodeStalls / cpu->baseStats.numCycles),
```
seanwang2046, line 248: does the "front end bound/frontend latency bound/frontend BW band" calculated every cycle or  calculated just once after the running? if calculated every cycle it may have bad impact on simulation speed. so I suggest we count the topdown PMU counters each cycle while calculate the topdown bounds just once after the running

src/cpu/o3/fetch.cc
```diff
@@ -1294,7 +1323,20 @@ Fetch::tick()
 
     toDecode->fetchStallReason = stallReason;
 
-    fetchStats.fetchStatusDist[fetchStatus[*tid_itr]]++;
+    // Intel TopDown method for measuring frontend bubbles
+    // Count unutilized issue slots when backend is not stalled (decode not stalled)
+    // For N-wide machine, if frontend supplies 0 instructions:
+    // - fetchBubbles += N (count total empty slots)
+    // - fetchBubbles_max += 1 (count occurrence of all slots being empty)
+    if (!stalls[*tid_itr].decode) { // backend not stalled
```
seanwang2046: does fetch bubble also count when flush bubble occurs? we expect that fetch bubble should not count when flush occurs, because flush penalty should be count as part of bad spec

src/cpu/o3/issue_queue.cc
```diff
@@ -764,6 +781,18 @@ Scheduler::issueAndSelect()
     for (auto it : issueQues) {
         it->issueToFu();
     }
+    if (instsToFu.size() < 4) {
```
seanwang2046: I suggest we use a macro to set the value of the FEW, not just type "4" here

src/cpu/o3/issue_queue.cc
```diff
@@ -623,6 +623,22 @@ Scheduler::SpecWakeupCompletion::description() const
...
 }
 
+Scheduler::SchedulerStats::SchedulerStats(statistics::Group* parent)
+  : statistics::Group(parent),
+    ADD_STAT(exec_stall_cycle, "SUM(OpsExecuted[= FEW])"),
+    ADD_STAT(memstall_any_load,
+        "Cycles with no uops executed and at least X in-flight load that is not completed yet"),
+    ADD_STAT(memstall_any_store,
+        "Cycles with few uops executed and no more stores can be issued"),
+    ADD_STAT(memstall_l1miss,
```
seanwang2046, line 636: the comment here need also modify to "least X"

src/cpu/o3/lsq.cc
```diff
@@ -1352,6 +1383,12 @@ LSQ::SingleDataRequest::recvTimingResp(PacketPtr pkt)
     // Dump inst num, request addr, and packet addr
     DPRINTF(LSQ, "Single Req::recvTimingResp: inst: %llu, pkt: %#lx\n", pkt->req->getReqInstSeqNum(),
             pkt->getAddr());
+    if (isLoad()) {
```
seanwang2046, line 1480: what does this part of code do here?

src/cpu/o3/commit.cc
tastynoob: The mispred branch still will commit

src/cpu/o3/lsq.cc
tastynoob: This is used for release the infilght load queue' entry

src/cpu/o3/issue_queue.cc
tastynoob: Yes

happy-lx (approved):

src/cpu/o3/lsq_unit.hh
```diff
@@ -289,6 +289,8 @@ class LSQUnit
     using LoadQueue = CircularQueue<LQEntry>;
     using StoreQueue = CircularQueue<SQEntry>;
 
+    std::vector<LSQRequest*> inflightLoads;
```
happy-lx, line 292: After [#265](https://github.com/OpenXiangShan/GEM5/pull/265) , once an load instruction is replayed due to a cache miss, the request at the corresponding position in the LQ will be set to as `nullptr`, and this miss request will not be tracked. Using `request` to determine if an instruction is inflight might lead to errors.

Maybe use `seqNum` to track inflightLoads can solve these errors

PR #264, opened 2025-01-08
slice stall and dcache write stall

tastynoob:


src/mem/cache/base.cc
```diff
@@ -2990,6 +2999,18 @@ BaseCache::CpuSidePort::tryTiming(PacketPtr pkt)
...
     }
+    int sliceidx = cache->getSliceIdx(pkt->getAddr());
+    if (sliceidx >= 0 && cache->cacheLevel != 1) {
+        if (cache->checkSLiceBusy(pkt, sliceidx)) {
+            //no more buffer
+            if (sendRetryEvent.scheduled()) {
+                owner.reschedule(sendRetryEvent, curTick());
+            } else {
+                owner.schedule(sendRetryEvent, curTick());
+            }
```
happy-lx, line 3010: Why is the retry packet sent at curTick?

tastynoob: It will inert this event at curcycle's end, so it's equivalent to schedule event at lastcycle's begin

src/cpu/o3/lsq_unit.cc
```diff
@@ -1404,16 +1404,20 @@ bool LSQUnit::insertStoreBuffer(Addr vaddr, Addr paddr, uint8_t* datas, uint64_t
...
         assert(storeBuffer.unsentSize() == 0);
         storeBufferFlushing = false;
         cpu->activityThisCycle();
+    }
+
+    // write request will stall one cycle
+    // so 2 cycle send one write request
+    if (lsq->getDcacheWriteStall()) {
+        lsq->setDcacheWriteStall(false);
         return;
```
happy-lx, line 1417: The Dcache requeset of this store may write DataSram after N cycles (4 if Hit, 10+/100+ if miss) arriving at Dcache, (I feel that the blocking should be done at that time instead of blocking directly in the next cycle?

happy-lx (approved):

PR #265, opened 2025-01-10
cpu-o3: Transform the lsqunit

happy-lx:
## Background
In order to align the behavior of the LSU module with the RTL, this PR makes the following modifications:
+ Uses TimeBuffer to visualize the load and store pipelines.
+ Distributes the operational details of the load and store pipelines across the corresponding pipeline stages.
+ Adds nuke replay operation to reduce RAW violations.
+ Aligns miss load replay behavior with RTL and adds early wake-up.

### 1. Pipeline Construction
Related commits:
+ [f5039874530fdd4ec7ca8097e13c8c3d20607e84](https://github.com/OpenXiangShan/GEM5/pull/265/commits/f5039874530fdd4ec7ca8097e13c8c3d20607e84)
+ [89fdfd0a589d5390e12d2c016fa99d6b59f4e1c1](https://github.com/OpenXiangShan/GEM5/pull/265/commits/89fdfd0a589d5390e12d2c016fa99d6b59f4e1c1)
+ [bfbf7c957c2a873aee9105664d1228233ceb4f92](https://github.com/OpenXiangShan/GEM5/pull/265/commits/bfbf7c957c2a873aee9105664d1228233ceb4f92)

**Original design:** Load/store operations are delayed by a certain number of cycles and then all operations (TLB, Cache lookup, exception check, etc.) are completed in one cycle using `executeLoad/executeStore`.

**New design:** TimeBuffer is set according to the corresponding pipeline stages. Instructions are first dispatched to TimeBuffer `s0`, and over time they pass through the stages of the pipeline to complete the corresponding operations.

<details>
  <summary>Code Details</summary>

  **Original design:**

  + The operation was delayed by `op_latency - 1` cycles before being sent to the corresponding function unit. https://github.com/OpenXiangShan/GEM5/blob/2d1995aa9820b5b26b68002e698960b628e072cb/src/cpu/o3/inst_queue.cc#L700-L708
  + After reaching the function unit, the load/store calls `executeLoad/executeStore` to complete the corresponding operations. https://github.com/OpenXiangShan/GEM5/blob/2d1995aa9820b5b26b68002e698960b628e072cb/src/cpu/o3/iew.cc#L1407-L1437
  + `executeLoad/executeStore` completes all operations. https://github.com/OpenXiangShan/GEM5/blob/2d1995aa9820b5b26b68002e698960b628e072cb/src/cpu/o3/lsq_unit.cc#L906-L979 https://github.com/OpenXiangShan/GEM5/blob/2d1995aa9820b5b26b68002e698960b628e072cb/src/cpu/o3/lsq_unit.cc#L1003-L1067

   **New design:**

  + The load/store operations are directly sent to the function unit. https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/inst_queue.cc#L711-L719
  + After reaching the function unit, load/store operations call `issueToLoadPipe/issueToStorePipe` to send the instructions to the `s0` of the load/store pipeline. `loadPipeSx`/`storePipeSx` is the corresponding pipeline timebuffer, `loadPipeSx[0]` means s0 stage of load pipeline.  https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.hh#L658-L682 https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/iew.cc#L1473-L1503 https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L1053-L1081
  + Over time, the instructions gradually enter the subsequent pipeline stages (`s1`, `s2`, etc.). The corresponding instructions are fetched from the load/store pipeline at each stage and the operations are executed. https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L1314-L1367 https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L1493-L1559
</details>

### 2. Nuke Replay
Related commit:
+ [bfbf7c957c2a873aee9105664d1228233ceb4f92](https://github.com/OpenXiangShan/GEM5/pull/265/commits/bfbf7c957c2a873aee9105664d1228233ceb4f92)

> Scenario:
> ```asm
> sw s1, 0(s0)
> lw s2, 0(s0)
> ```
> The load needs to forward data from the store, and both load and store are issued simultaneously from the IQ into the pipeline, arriving at `s1` in the load/store pipeline at the same time.

**Original design:** If the load executes first, since the store in the same cycle hasn't finished (and hasn't updated the SQ), the load can't detect the dependency and thus can't forward data from the store. Then the store in the same cycle executes, a RAW violation will be detected, causing the pipeline to be flushed.

**What about RTL behavior:** The described situation does not occur in the RTL because the store instruction in the s1 pipeline stage (the cycle before the SQ is updated) checks whether there is a load instruction on the load pipeline with the same address and data that has not forwarded from it, and causes that load instruction to be replayed.

**New design:** When the load reaches `s1` or `s2`, it checks whether there are stores still in `s1` that haven't executed and match the address. If so, the load will be replayed. This is called the pipeline nuke replay. After the store in `s1` completes, the address and data are updated to the SQ, allowing the replayed load to correctly forward data without causing a RAW violation.

> **You can use `EnablePipeNukeCheck` to control whether to enable this feature. In addition, if you want to use this feature, make sure `EnableLdMissReplay` is set to `True`.**

<details>
  <summary>Code Details</summary>

**RTL design:**
  + store instructions at store pipeline `s1` query load pipeline `s1/s2`, causing the matched load to be replayed https://github.com/OpenXiangShan/XiangShan/blob/0051450372ae5a03ce9d36afdbdd34b9a19f4785/src/main/scala/xiangshan/mem/pipeline/LoadUnit.scala#L971-L995 https://github.com/OpenXiangShan/XiangShan/blob/0051450372ae5a03ce9d36afdbdd34b9a19f4785/src/main/scala/xiangshan/mem/pipeline/LoadUnit.scala#L1228-L1248 https://github.com/OpenXiangShan/XiangShan/blob/0051450372ae5a03ce9d36afdbdd34b9a19f4785/src/main/scala/xiangshan/mem/pipeline/LoadUnit.scala#L1357-L1381

 **Gem5 New design:**

  + The `pipeLineNukeCheck` is where the load in `s1` checks for pipeline nuke. https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L1159-L1179
  + The `pipeLineNukeCheck` for load in `s2` checks for pipeline nuke. https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L1250-L1258
  + Check logic. https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L743-L762
  + If the load in `s1` or `s2` triggers a pipeline nuke, a fast load replay is performed in `s2`. https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L1273-L1282
  + If the load needs to perform a pipeline nuke replay, the pipeline flag of this load will be marked as `Nuke`, and the RAW violation check will skip this load. https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L955-L961
</details>

### 3. Miss Load Replay  
Commits related to this modification:  
+ [ee53d22929c24e56c386efc13190b74c90f8997c](https://github.com/OpenXiangShan/GEM5/pull/265/commits/ee53d22929c24e56c386efc13190b74c90f8997c)  
+ [c3cd30088a4d5a57a47b54a43c53473436b5650c](https://github.com/OpenXiangShan/GEM5/pull/265/commits/c3cd30088a4d5a57a47b54a43c53473436b5650c)  

**Original design:** A miss load directly writes back after receiving the `TimingResp`, with no limit on the number and no re-entering pipeline behavior.

**New design:** A miss load receives a custom `Hint` signal two cycles before receiving the `TimingResp`, which wakes up the relevant instructions and reissues them into the pipeline. The relevant data becomes stable in the `bus` after `TimingResp` reaching the LSU, and the reissued load will forward data from the `bus`. If data cannot be forwarded, it will access the cache again. After the cache refilled the data to itself completely, a `Bus_Clear` request is sent to the LSU to clear the corresponding data in the `bus`. 

> **You can use `EnableLdMissReplay` to control whether to enable this feature.**

<details>  
  <summary>Code details</summary>  

  **Original design:**

  + A miss load directly calls `completeDataAccess` for write-back after receiving the `TimingResp` https://github.com/OpenXiangShan/GEM5/blob/2d1995aa9820b5b26b68002e698960b628e072cb/src/cpu/o3/lsq.cc#L1349-L1362

  **New design:**

  + A miss load enters a `cache miss load replayQueue` https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L1259-L1271
  + When the refill data reaches the bus between L2 and L1, a `Hint` signal is returned a few cycles ahead of the `TimingResp` https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/mem/coherent_xbar.cc#L518-L522
  + After the `Hint` reaches L1, it queries the MSHR and sends a wake-up signal to the relevant load requests in the cache block. The loads that receive the wake-up signal are awakened from the replayQueue https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/mem/cache/cache.cc#L963-L999 https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq.cc#L1524-L1537
  + The actual `TimingResp` arrives a few cycles after the `Hint` and stabilizes the data on the bus https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq.cc#L1434-L1475
  + The reissued load checks the bus before accessing the cache. If data is available on the bus, it directly uses the data from the bus (via `forwardFrmBus(load_inst, request)`), and accesses the cache only if no data is found on the bus. https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L2995-L3023
  + Sure, it is also possible that the Hint is sent too early, causing the reissued load reaching the pipeline and find that the TimingResp has not yet arrived. In this case, the load will speculatively wait for one more cycle, i.e., wait until s2. If the TimingResp arrives, it can forward data from the bus in s2 and then write back. If s2 still hasn't received the TimingResp, it will re-enter the cache miss replay queue.(**Note: this scenario will be very rare.**) https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L2995-L3002 https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq_unit.cc#L1221-L1244
  + After the cache refill data is ready within the cache, a `Bus_Clear` is sent to the LSU to clear the relevant data from the bus https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/mem/cache/base.cc#L2041-L2052 https://github.com/OpenXiangShan/GEM5/blob/cf1302416f681eded0b2ba9f229dc27434cdfb83/src/cpu/o3/lsq.cc#L586-L602
</details>

### 4. Misc
During the alignment process, some functionality correctness issues and minor misalignments were also discovered, and they are all documented in this section.

#### 4.1 fence opType
Commits related to this modification:  
+ [5e118d7bfbffc7d18a4ad52b73330e61221e7713](https://github.com/OpenXiangShan/GEM5/pull/265/commits/5e118d7bfbffc7d18a4ad52b73330e61221e7713)  

**Original design:**

the fence instruction will be dispatched to mem's dispatchQueue, but its opType is `No_OpClass`, which will cause it to wait for the integer issue queue IQ2(IntMisc) to have free items before it can continue execution.
If the subsequent instructions of the fence instruction occupy the intIQ2, the fence cannot be executed and cpu deadlocks.

**new design:** 

Change the opType of the fence instruction to `MemReadOp` to prevent this situation (in fact, the fence will not be dispatched to IQ)

#### 4.2 LRSC
Commits related to this modification:  
+ [af375fa600564b38392501242e9aa1ddd19d2b56](https://github.com/OpenXiangShan/GEM5/pull/265/commits/af375fa600564b38392501242e9aa1ddd19d2b56)  

LR Instruction can be executed speculatively, which causes RAW violations in some corner cases.
**To reduce complexity and ensure consistency with the RTL design, avoid speculative execution for the LR, and strictly maintain order.**

#### 4.3 store writeback
Commits related to this modification:  
+ [3891ceb1bcc9cb2245b02c61e1fad3ecbefc6758](https://github.com/OpenXiangShan/GEM5/pull/265/commits/3891ceb1bcc9cb2245b02c61e1fad3ecbefc6758)

Previously, the store was written back in s2, but it is written back in s4 in RTL design. So align the timing of this operation. **The write-back timing has a significant impact on high IPC programs like hmmer (~13%), with earlier write-back being more beneficial for improving performance.**

> **You can set `StoreWbStage` to 2 or 4 or N to control which stage the store is written back.**

### DIFF
--- a/src/cpu/o3/lsq.cc
+++ b/src/cpu/o3/lsq.cc
@@ -88,6 +88,7 @@ LSQ::LSQ(CPU *cpu_ptr, IEW *iew_ptr, const BaseO3CPUParams &params)
       enableBankConflictCheck(params.BankConflictCheck),
       _enableLdMissReplay(params.EnableLdMissReplay),
       _enablePipeNukeCheck(params.EnablePipeNukeCheck),
+      _storeWbStage(params.StoreWbStage),
       waitingForStaleTranslation(false),
       staleTranslationWaitTxnId(0),
       lsqPolicy(params.smtLSQPolicy),
@@ -104,6 +105,7 @@ LSQ::LSQ(CPU *cpu_ptr, IEW *iew_ptr, const BaseO3CPUParams &params)
     if (!_enableLdMissReplay && _enablePipeNukeCheck) {
         panic("LSQ can not support pipeline nuke replay when EnableLdMissReplay is False");
     }
+    assert(_storeWbStage >= 2 && _storeWbStage <= 4);
 
     //**********************************************
     //************ Handle SMT Parameters ***********
--- a/src/cpu/o3/lsq.cc
+++ b/src/cpu/o3/lsq.cc
@@ -43,6 +43,7 @@
 
 #include <algorithm>
 #include <csignal>
+#include <cstdint>
 #include <list>
 #include <string>
 
@@ -51,6 +52,7 @@
 #include "base/trace.hh"
 #include "cpu/o3/cpu.hh"
 #include "cpu/o3/dyn_inst.hh"
+#include "cpu/o3/dyn_inst_ptr.hh"
 #include "cpu/o3/iew.hh"
 #include "cpu/o3/limits.hh"
 #include "debug/Drain.hh"
@@ -323,19 +325,11 @@ LSQ::executePipeSx()
 }
 
 Fault
-LSQ::executeLoad(const DynInstPtr &inst)
+LSQ::executeAmo(const DynInstPtr &inst)
 {
     ThreadID tid = inst->threadNumber;
 
-    return thread[tid].executeLoad(inst);
-}
-
-Fault
-LSQ::executeStore(const DynInstPtr &inst)
-{
-    ThreadID tid = inst->threadNumber;
-
-    return thread[tid].executeStore(inst);
+    return thread[tid].executeAmo(inst);
 }
 
 void
@@ -565,8 +559,12 @@ LSQ::recvFunctionalCustomSignal(PacketPtr pkt, int sig)
     LSQRequest *request = dynamic_cast<LSQRequest*>(pkt->getPrimarySenderState());
     panic_if(!request, "Got packet back with unknown sender state\n");
     if (sig == DcacheRespType::Miss) {
-        // notify cache miss
-        iewStage->loadCancel(request->instruction());
+        if (request->instruction()->isLoad()) {
+            // notify cache miss
+            iewStage->loadCancel(request->instruction());
+            // set cache miss flag in pipeline
+            thread[request->_port.lsqID].setFlagInPipeLine(request->instruction(), LdStFlags::CacheMiss);
+        }
     } else {
         panic("unsupported sig %d in recvFunctionalCustomSignal\n", sig);
     }
@@ -958,6 +956,14 @@ LSQ::pushRequest(const DynInstPtr& inst, bool isLoad, uint8_t *data,
         request->initiateTranslation();
     }
 
+    if (!isLoad && !isAtomic) {
+        // store inst temporally saves its data in memData
+        inst->memData = new uint8_t[size];
+        memcpy(inst->memData, data, size);
+    }
+
+    inst->effSize = size;
+
     if (!isLoad && !inst->isVector() && size > 1 && addr % size != 0) {
         warn( "Store misaligned: size: %u, Addr: %#lx, code: %d\n", size,
             addr, RiscvISA::ExceptionCode::STORE_ADDR_MISALIGNED);
@@ -967,7 +973,7 @@ LSQ::pushRequest(const DynInstPtr& inst, bool isLoad, uint8_t *data,
     }
 
     /* This is the place were instructions get the effAddr. */
-    if (request->isTranslationComplete()) {
+    if (inst->isAtomic() && request->isTranslationComplete()) {
         if (request->isMemAccessRequired()) {
             inst->effAddr = request->getVaddr();
             inst->effSize = size;
--- a/src/cpu/o3/lsq.cc
+++ b/src/cpu/o3/lsq.cc
@@ -558,13 +558,26 @@ LSQ::recvFunctionalCustomSignal(PacketPtr pkt, int sig)
 
     LSQRequest *request = dynamic_cast<LSQRequest*>(pkt->getPrimarySenderState());
     panic_if(!request, "Got packet back with unknown sender state\n");
-    if (sig == DcacheRespType::Miss) {
-        if (request->instruction()->isLoad()) {
-            // notify cache miss
-            iewStage->loadCancel(request->instruction());
+    if (sig == DcacheRespType::Miss || sig == DcacheRespType::Block_Not_Ready) {
+        DPRINTF(LSQ, "recvFunctionalCustomSignal: Resp type: %d, [sn:%ld], lqidx: %ld\n",
+                sig, request->instruction()->seqNum, request->instruction()->lqIdx);
+        if (request->mainReq()->isLLSC() || request->mainReq()->isUncacheable()) {
+            // do not replay Amo/Uncache Load
+            DPRINTF(LSQ, "Recv Amo/Uncache Load: [sn:%ld], No Need to Replay\n",
+                    request->instruction()->seqNum);
+        } else {
+            // clear state in this instruction
+            request->instruction()->cacheRefilledAfterMiss(false);
+            request->instruction()->effAddrValid(false);
+            // clear request in loadQueue
+            thread[request->_port.lsqID].loadQueue[request->instruction()->lqIdx].setRequest(nullptr);
             // set cache miss flag in pipeline
             thread[request->_port.lsqID].setFlagInPipeLine(request->instruction(), LdStFlags::CacheMiss);
+            // insert to missed load replay queue
+            iewStage->cacheMissLdReplay(request->instruction());
         }
+        // cancel subsequent dependent insts of this load
+        iewStage->loadCancel(request->instruction());
     } else {
         panic("unsupported sig %d in recvFunctionalCustomSignal\n", sig);
     }
@@ -1397,21 +1410,32 @@ LSQ::SbufferRequest::recvTimingResp(PacketPtr pkt)
 bool
 LSQ::SingleDataRequest::recvTimingResp(PacketPtr pkt)
 {
+    LSQRequest *request = dynamic_cast<LSQRequest*>(pkt->senderState);
+    bool isNormalLd = isLoad() && !request->mainReq()->isLLSC() && !request->mainReq()->isUncacheable();
     // Dump inst num, request addr, and packet addr
-    DPRINTF(LSQ, "Single Req::recvTimingResp: inst: %llu, pkt: %#lx\n", pkt->req->getReqInstSeqNum(),
-            pkt->getAddr());
+    DPRINTF(LSQ, "Single Req::recvTimingResp: inst: %llu, pkt: %#lx, isLoad: %d, "
+                "isLLSC: %d, isUncache: %d, isCacheSatisfied: %d\n",
+                pkt->req->getReqInstSeqNum(), pkt->getAddr(), isLoad(), request->mainReq()->isLLSC(),
+                request->mainReq()->isUncacheable(), pkt->cacheSatisfied);
     assert(_numOutstandingPackets == 1);
-    flags.set(Flag::Complete);
-    assert(pkt == _packets.front());
-    forward();
-    _port.completeDataAccess(pkt);
-    _hasStaleTranslation = false;
+    if (isNormalLd && !pkt->cacheSatisfied) {
+        // Data in Dcache is ready, wake up missed load in replay queue
+        LSQRequest::_inst->cacheRefilledAfterMiss(true);
+        discard();
+    } else {
+        flags.set(Flag::Complete);
+        assert(pkt == _packets.front());
+        forward();
+        _port.completeDataAccess(pkt);
+        _hasStaleTranslation = false;
+    }
     return true;
 }
 
 bool
 LSQ::SplitDataRequest::recvTimingResp(PacketPtr pkt)
 {
+    LSQRequest *request = dynamic_cast<LSQRequest*>(pkt->senderState);
     DPRINTF(LSQ, "Spilt Req::recvTimingResp: inst: %llu, pkt: %#lx\n", pkt->req->getReqInstSeqNum(),
             pkt->getAddr());
     uint32_t pktIdx = 0;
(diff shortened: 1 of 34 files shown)


jueshiwenli (approved):

src/cpu/o3/lsq.cc
```diff
@@ -121,9 +135,10 @@ LSQ::LSQ(CPU *cpu_ptr, IEW *iew_ptr, const BaseO3CPUParams &params)
     }
 
     thread.reserve(numThreads);
+    // TODO: Parameterize the load/store pipeline stages
```
tastynoob, line 138: parameterize this

src/cpu/o3/lsq.cc
```diff
@@ -1282,7 +1365,9 @@ LSQ::LSQRequest::~LSQRequest()
         std::raise(SIGINT);
     }
     assert(!isAnyOutstandingRequest());
-    _inst->savedRequest = nullptr;
+    if (_inst->savedRequest == this) {
```
tastynoob, line 1369: why need this condition

src/cpu/o3/lsq_unit.cc
```diff
@@ -947,35 +1156,413 @@ LSQUnit::executeLoad(const DynInstPtr &inst)
...
+        if (enableStorePrefetchTrain) {
+            triggerStorePFTrain(store_idx);
+        }
+    }
+
+    return checkViolations(loadIt, inst);
+}
+
+Fault
+LSQUnit::storePipeS2(const DynInstPtr &inst, std::bitset<LdStFlagNum> &flag)
```
tastynoob: Why so many identical function?
Can be simplified

src/cpu/o3/lsq_unit.cc
```diff
@@ -1973,6 +2548,40 @@ LSQUnit::recvRetry()
     }
 }
 
+void
+LSQUnit::dumpLoadPipe()
+{
+    DPRINTF(LSQUnit, "Dumping LoadPipe:\n");
```
tastynoob: use:
if (debug::LSQUnit)

src/cpu/o3/lsq_unit.cc
```diff
@@ -1973,6 +2548,40 @@ LSQUnit::recvRetry()
...
+                    getLdStFlagStr(loadPipeSx[i]->flags[j])
+            );
+        }
+    }
+}
+
+void
+LSQUnit::dumpStorePipe()
+{
+    DPRINTF(LSQUnit, "Dumping StorePipe:\n");
```
tastynoob: Same as above

src/cpu/o3/lsq_unit.hh
```diff
@@ -586,6 +655,35 @@ class LSQUnit
     /** The load queue. */
     LoadQueue loadQueue;
 
+    /** Struct that defines the information passed through Load Pipeline. */
+    struct LoadPipeStruct
+    {
+        int size;
+
+        DynInstPtr insts[MaxWidth];
+        std::bitset<LdStFlagNum> flags[MaxWidth];
```
tastynoob, line 662: I suggest put the ldsrflags into dyninst

src/cpu/o3/lsq_unit.hh
```diff
@@ -586,6 +655,35 @@ class LSQUnit
...
+    /** Each stage in load pipeline. loadPipeSx[0] means load pipe S0 */
+    std::vector<TimeBuffer<LoadPipeStruct>::wire> loadPipeSx;
+
+    /** Struct that defines the information passed through Store Pipeline. */
+    struct StorePipeStruct
+    {
+        int size;
+
+        DynInstPtr insts[MaxWidth];
+        std::bitset<LdStFlagNum> flags[MaxWidth];
```
tastynoob, line 675: same as above

jensen-yan (approved):

tastynoob (approved):

src/cpu/o3/lsq_unit.hh
happy-lx: In the current design, where a single instruction may appear in both s0 and s3 of the pipeline (fast replay), I feel storing the state in the TimeBuffer would be easier to manage.

src/cpu/o3/lsq_unit.cc
happy-lx: Yes, using `debug::LSQUnit` is much better for simulation speed, I will fix it.

src/cpu/o3/lsq_unit.cc
happy-lx: Yes. These `storePipeS2`, `S3`, and `S4` do not perform any operations and are indeed redundant. I will replace them with a new function.

src/cpu/o3/lsq.cc
happy-lx: In the current design, after a load instruction experiences a miss for the first time, it does not continue to track the corresponding request. 

Once the miss request retrieves the data, refills the dcache, and returns a response to the LSU, it will discard itself. 

When the missed load is replayed to the pipeline, a new request will be generated(the goal is to ensure it re-accesses the DTLB and dcache when necessary). **This approach indeed breaks the rule that each instruction is correspond to only one request. If the request management is not handled properly, it could potentially lead to functional issues.**

However, in order to implement early wake-up and align with the RTL, I feel this approach is actually more friendly in terms of implementation.

src/cpu/o3/lsq.cc
happy-lx: OK, I will fix this

PR #266, opened 2025-01-11
cpu-o3: delete pending wake events if cancel

happy-lx:
I encountered a bug with the following scenario:

```asm
fld fa4, 0(a0)
fmul fa5, fa5, fa4
fst fa5, 0(a0)
```

Bug scenario:
1. After `fld` is scheduled, it speculatively wakes up `fmul`.
2. After `fmul` is scheduled, a speculative wake-up event (`event1`) for `fst` is generated, and the actual wake-up occurs after a few cycles of delay.
3. `fld` is cancelled, which cancels the speculative wake-up for `fmul`, setting `fmul` as canceled, **but `event1` is not cleared.**
4. When `fld` is reissued, it completes and deterministically wakes up `fmul`, when `fmul` schedules, **`fmul->canceled()` is false**. At the same tick, `event1` starts to execute and wakes up `fst`.
5. When `fst` is scheduled, it is unable to obtain data from the bypass network(because the wake-up relationship between `fmul` and `fst` is broken).

The root cause of this bug is that **the speculative wake-up event was not canceled along with the instruction when it was canceled.**

This PR does modifications that when an instruction is canceled, its pending speculative wake-up events are also canceled.

Change-Id: Ib4c325f42de5874c4cc8c40c4389b9c8985df9f3

### DIFF
--- a/src/cpu/o3/issue_queue.cc
+++ b/src/cpu/o3/issue_queue.cc
@@ -215,7 +215,7 @@ IssueQue::checkScoreboard(const DynInstPtr& inst)
         if (!scheduler->bypassScoreboard[src->flatIndex()]) [[unlikely]] {
             auto dst_inst = scheduler->getInstByDstReg(src->flatIndex());
             if (!dst_inst || !dst_inst->isLoad()) {
-                panic("dst is not load");
+                panic("dst[sn:%llu] is not load", dst_inst->seqNum);
             }
             scheduler->loadCancel(dst_inst);
             DPRINTF(Schedule, "[sn:%llu] %s can't get data from bypassNetwork, dst inst: %s\n", inst->seqNum,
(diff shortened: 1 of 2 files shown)


src/cpu/o3/issue_queue.cc
```diff
@@ -606,15 +606,16 @@ IssueQue::doSquash(const InstSeqNum seqNum)
...
+Scheduler::SpecWakeupCompletion::SpecWakeupCompletion(const DynInstPtr& inst, IssueQue* to, Scheduler* scheduler)
+    : Event(Stat_Event_Pri, AutoDelete), inst(inst), scheduler(scheduler), to_issue_queue(to)
 {
 }
 
 void
 Scheduler::SpecWakeupCompletion::process()
 {
     to_issue_queue->wakeUpDependents(inst, true);
+    scheduler->specWakeEvents.erase(this);
```
tastynoob: Could we save the iterator pointer to delere it faster

src/cpu/o3/issue_queue.cc
```diff
@@ -983,6 +986,16 @@ Scheduler::loadCancel(const DynInstPtr& inst)
...
+        // clear pending wake events scheduled by top
+        for (auto it = specWakeEvents.begin(); it != specWakeEvents.end();) {
+            auto e = *it;
+            if (e->inst == top) {
+                cpu->deschedule(e);
+                it = specWakeEvents.erase(it);
+            } else {
+                it++;
+            }
+        }
```
tastynoob, line 995: Same as above

src/cpu/o3/issue_queue.hh
```diff
@@ -193,11 +194,12 @@ class Scheduler : public SimObject
     friend class IssueQue;
     class SpecWakeupCompletion : public Event
     {
+      public:
         DynInstPtr inst;
+        Scheduler* scheduler;
```
tastynoob: Here only to save specWakeEvents's pointer

tastynoob (approved):

PR #268, opened 2025-01-15
cpu,difftest: use memcpy_init when init difftest

eastonman:
- Add `memcpy_init` function to NemuProxy class
- Use `memcpy_init` instead of `memcpy` when memory deduplication is disabled

This commit use memcpy_init instead of memcpy when init difftest. This is corresponded to NEMU commit
https://github.com/OpenXiangShan/NEMU/commit/3cc6d1515c06efdde9574a04ef96e664ce2c5ac3, which moves `nemu_large_memcpy` from memcpy to memcpy_init. This change fix the problem of NEMU side pmem is fully allocated on init.


jensen-yan:
I think you should update nemu-ref-so in CI path, both in small/large server(open/node).

eastonman:
will close this PR and open a new one combined with #269. Then update the ref NEMU so.

PR #274, opened 2025-01-21
cpu-o3: fix a memory leak problem

happy-lx:
After the cache refills the data, it sends a custom `Bus_Clear` packet to notify the LSU. This should only be sent by the Dcache, and neither L2 nor L3 should send this packet.

Calling `recvFunctionalCustomSignal` on the xbar does nothing, causing the custom packet to be lost, leading to a memory leak.

Thanks to @tastynoob finding this problem

Change-Id: Ie735660b66536908633daa6c25d76b7127add0b9

### DIFF
--- a/src/cpu/o3/lsq.cc
+++ b/src/cpu/o3/lsq.cc
@@ -633,7 +633,6 @@ LSQ::recvFunctionalCustomSignal(PacketPtr pkt, int sig)
                 it++;
             }
         }
-        delete pkt;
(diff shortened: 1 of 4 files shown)


tastynoob (approved):

tastynoob:
Should delete pkt only when sig was hint or bus_clear

happy-lx:
> Should delete pkt only when sig was hint or bus_clear

Yes I will fix it, the packet used by the hint is the packet from the `mshr`'s target, which may not need to be deleted after the `customEvent` is completed. I can make a distinction between them.

PR #292, opened 2025-02-20
cpu-o3: implement compress rob

tastynoob:


happy-lx:
Is the width of regFile freeing and renaming table modification also based on a maximum of 8 groups / cycle?

jueshiwenli (approved):

Issue #215, opened 2024-11-28
Idea about the structure of an ideal predictor for architect exploration


zybzzz:
I am currently working on the design of a value predictor. I have understood the use of **arch_db(trace)** and have some new ideas. For the value predictor, we use arch_db to record runtime traces. We then adjust our designed value predictor based on these traces, attempting to optimize it—for example, achieving an accuracy rate above 95%. After that, we perform SPEC06 benchmarking to analyze the gains we've obtained.

In this process, the final benefits we get from the value predictor are **unknown**; that is, we don't know the **ideal gains** we would achieve if all instructions were predicted successfully. For example, when we optimize branch prediction, we know that better branch prediction can improve the instruction fetch bandwidth, thereby **absolutely** enhancing processor performance. Similarly, when we optimize cache prefetching, we know that effective prefetching can definitely reduce memory access overhead, thus **absolutely** improving processor performance. **However**, with value prediction technology, we **cannot be certain** that it will yield better performance even when it works well. Simply put, when we predict ```load``` instructions, even if the prediction is correct, the correct result might be wasted due to violations of load/store memory access ordering. Similarly, when predicting ```add``` instructions, the predicted addition might not be an instruction affected by true dependencies in the out-of-order execution backend; even if the prediction is correct, it **may not significantly improve** performance and might instead incur the overhead of incorrect predictions. **I believe all these reasons boil down to the fact that optimizing backend execution in an out-of-order processor involves too many considerations.**

Based on this, I came up with an idea: we can try to first obtain the best-case benefits of value prediction technology. For example, we can use a emulator like NEMU that directly provides absolutely correct results when predicting instructions. This effectively simulates a value predictor with 100% accuracy. We can then perform benchmarking evaluations on such a value predictor to know in advance the **maximum gains we can achieve before we start trace tuning**, giving us more confidence to proceed with optimization. If the designed interface is sufficiently excellent and general, I believe this approach can help us quickly perform design space exploration and rapidly analyze the benefits that our design can bring.

In comparison, both tuning the value predictor to optimal performance using traces and constructing an ideal predictor with 100% accuracy **aim to determine how much benefit our design can achieve under optimal conditions**. For trace-based analysis, the difficulty lies in manually analyzing the traces and tuning based on them. For the ideal predictor, the challenge is in designing a NEMU that provides 100% correct values, integrating it into gem5, and ensuring that the instruction streams between the NEMU and gem5 are synchronized.

[...]

PR #297, opened 2025-02-20
ideal model and value predictor support

zybzzz:
related: #215 #188 OpenXiangShan/NEMU#806

Added support for the ideal model and value prediction in this PR.

1. **Support for the Ideal Model**: The ideal model can now be applied to int add-type instructions and scalar load operations for upper-bound performance analysis of value prediction.
2. **Support for Value Prediction**: A basic implementation of value prediction has been provided. Since the previous value prediction implementation could not be toggled on or off, the related interfaces have been revised accordingly.

Through my testing on the server, both of these additions do not affect the execution of xs-gem5 when disabled.

---

The upstream merge of nemu will be a slow process, as the upstream nemu differs significantly from the version currently used for difftest, requiring time for integration. If this submission is accepted, I will first provide a precompiled version in the ready-to-run repository.


src/cpu/valuepred/enhanced_stride.cc
```diff
@@ -0,0 +1,569 @@
...
+
+
+    // call back to dump stats
+    registerExitCallback([this]() {
+        auto outhandle = simout.create("EStride.txt", false, true);
+
+        int ESSize = ways * entryCounts * (tagWidth + logMaxConfidence + strideWidth + 64 + 2 + 1);
+        *outhandle->stream() << "EStride Size: " << (1.0 * ESSize) / (8 * 1024)  << "KB" << std::endl;
+
+        *outhandle->stream() << "VPcorrect: " << stats.VPcorrected.value() << std::endl;
```
tastynoob: Why here dump again?

src/cpu/o3/fetch.cc
```diff
@@ -1936,17 +2088,119 @@ Fetch::fetch(bool &status_change)
...
+                // the value prediction is often started after the fetch phase.
+
+                // This simulate valuePredict in fetch the instructions
+                // no instruction information, every instructions take into valuePredictor.
+                valuepred::VPPredMetaData* vpPredMetaData =
+                        dynamic_cast<valuepred::ESPredMetaData *>(valuepred::VPDataStructFactory::
+                                                                        buildPredMetaData(ValuePredType::EStride));
+                vpPredMetaData->pc = instruction->getPC();
+                vpPredMetaData->seq_no = instruction->seqNum;
+                instruction->vpResult = valuePredictor->valuePredict(vpPredMetaData);
```
tastynoob, line 2191: Should it do predict at rename?
Allocate new physical register and remap to the predicted value.

src/cpu/o3/rename.cc
```diff
@@ -1270,6 +1276,52 @@ Rename::renameDestRegs(const DynInstPtr &inst, ThreadID tid)
                             rename_result.second);
 
         ++stats.renamedOperands;
+
+        if ((cpu->idealModelEnabled() && cpu->idealModelConfig->isValuePredSupported()) ||
```
tastynoob, line 1280: This block seems can be optimized.

src/cpu/valuepred/enhanced_stride.cc
zybzzz: > Why here dump again?

This part of the code can indeed be removed. I will modify it later, as it was used during my debugging.

src/cpu/o3/fetch.cc
zybzzz: > Should it do predict at rename? Allocate new physical register and remap to the predicted value.

In fact, the predicted values are indeed used in the Rename stage. What is being simulated here is a scenario where value prediction occurs when the instruction is fetched in the Fetch stage, and the predicted value is utilized in the Rename stage—essentially frontend prediction with backend consumption.

From a microarchitectural implementation perspective, accessing the value predictor in the Fetch stage instead of the Rename stage allows the predictor to obtain incoming instructions earlier. This leverages the latency between the frontend and backend to hide the overhead of value prediction.

src/cpu/o3/rename.cc
zybzzz: > This block seems can be optimized.

Currently, there are two scenarios where value prediction is performed:

1. **When the ideal model is enabled**
2. **When the value predictor is used independently**

The if condition here is used to distinguish between these two cases. The implementation of the ideal model is incompatible with the actual value predictor, which is why this "if" is necessary.

Furthermore, the strict checks on which instructions and registers are eligible for value prediction are currently unavoidable. **This is because the current implementation only supports value prediction for arithmetic instructions and standard loads, necessitating these strict checks on instructions and registers.**

zybzzz:
Due to the lack of consideration for the details in the instruction fetch stage, the ideal model still has some bugs that are currently being fixed.

zybzzz:
The issue is caused by the loop buffer, and I need some time to resolve it. Currently, XS-GEM5 is undergoing large-scale KunMingHu alignment, so I think the merge of this part of the code can be postponed until after the alignment.

@tastynoob @jensen-yan

eastonman:
I suggest just disabling the loop buffer for now. The loop buffer is very likely to be rewritten in the future.

zybzzz:
> I suggest just disabling the loop buffer for now. The loop buffer is very likely to be rewritten in the future.

Yes, I am currently using the ideal model when disabling the loop buffer. I will take this into account in the future merge.

PR #327, opened 2025-03-12
mem: align l2 bop config with rtl

happy-lx:
Change-Id: I19424ee4ce039abaedee09e06ca248609f7149e0


jueshiwenli (approved):

tastynoob (approved):

configs/common/PrefetcherConfig.py
```diff
@@ -64,6 +65,8 @@ def create_prefetcher(cpu, cache_level, options):
             prefetcher.enable_cmc = True
```
tastynoob: CMC on GEM5 should be disable

tastynoob (approved):

configs/common/PrefetcherConfig.py
happy-lx: okok

PR #342, opened 2025-03-18
util: add compile_command.json support

eastonman:
also deal with old version scons which does not support compilation_db tool

Change-Id: I206b6d7c9a9ac90199a4d973a19f4d4a416dc501

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ llvm-pgo/
 *nemu*
 ready-to-run/
 *.out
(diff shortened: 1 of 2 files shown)


eastonman:
This is derived from https://github.com/gem5/gem5/pull/1942. A manual pick is required because of quite a large divergence between xs-gem5 and mainline gem5.

jensen-yan (approved):

PR #374, opened 2025-05-12
revamped uBTB

boran-xu:
New, 32 entry uBTB that works with aBTB to produce a combined prediction at S1.
- its content is updated during BPU prediction pipeline
- it stores only one entry per FB (the first entry that it predict to jump)
32 entry uBTB + 1k entry aBTB is the default configuration if kmhv3-ideal is turned on and DecoupledBPUwithBTB is chosen as the BPU type.


eastonman (dismissed):
LGTM

jensen-yan (dismissed):

jensen-yan (approved):

PR #376, opened 2025-05-14
misc: add document to support website

jensen-yan:
add basic GEM5 chinese document skip-checks:true

Change-Id: I88d60471b1363b25fa09bce27d632cf7e582d90f


jensen-yan:
https://xs-gem5.readthedocs.io/zh-cn/latest/

PR #387, opened 2025-05-20
misc: add basic support for se mode[WIP][skip ci]

jensen-yan:
need to support difftest

Change-Id: Ib49f176ba1dbd71feebb423b6a3c47bdb78f82a7

### DIFF
--- a/configs/example/se.py
+++ b/configs/example/se.py
@@ -63,6 +63,7 @@ from common import MemConfig
 from common.FileSystemConfig import config_filesystem
 from common.Caches import *
 from common.cpu2000 import *
+from common.FUScheduler import DefaultScheduler
 
 def get_processes(args):
     """Interprets provided args and returns a list of processes"""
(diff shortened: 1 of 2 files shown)


jensen-yan:
nemu could only run raw binary, could not run elf, 
maybe we need to add difftest for qemu-riscv64, or do not support for se mode.

configs/example/se.py
```diff
@@ -237,6 +241,11 @@ def get_processes(args):
             ObjectList.indirect_bp_list.get(args.indirect_bp_type)
         system.cpu[i].branchPred.indirectBranchPred = indirectBPClass()
 
+    # For RISCV, configure MMU checker if needed
+    if buildEnv['TARGET_ISA'] == 'riscv':
+        system.cpu[i].mmu.pma_checker = PMAChecker(
```
cyyself: Why do we need this for the SE mode?

jensen-yan: I have deleted it, it is not necessary for se mode, thanks.

boran-xu (approved):

tastynoob (approved):

PR #389, opened 2025-05-21
configs: update pht-pf-level default and enhance prefetcher configuration

zephyrols:
Using act_entries=64 and pht_pf_level=2 provide a hyper performance than act_entries=32 and pht_pf_level=1 in solo L1 sms prefetcher. New Params solve the problem of negative optimization on omnetpp.

fca42381-449f-4859-ad56-7ed2cce8fd28

I also tried some other parameters, about (act_entries, pht_entries, pht_assoc, pht_pf_level). The results are as follows. From this, it can be observed that the common features of high-score configuration are a greater number of act entries and pht prefetching that spans multiple layers to the L2.
image


jensen-yan:
please give a brief reason of this PR, give desired performance stats if possible.

jensen-yan (approved):

PR #397, opened 2025-05-22
Tage use phr

boran-xu:


jensen-yan:
这个PR中包含了MGSC 的修改，相对于MGSC 19.715 的性能，将tage 的GHR 修改为PHR后，性能降低为19.693， 影响不大，修改PHR 还比较合理，之后可以考虑把target 也存入PHR 中来提升性能。              
This PR includes the modification of MGSC. Compared with the performance of MGSC 19.715, after changing the GHR of tag to PHR, the performance is reduced to 19.693, which has little impact. Modifying PHR is reasonable. In the future, we can consider storing the target in PHR to improve performance.

src/cpu/pred/btb/btb_tage.cc
```diff
@@ -31,6 +31,7 @@ enableSC(p.enableSC),
 tageStats(this, p.numPredictors)
 {
     DPRINTF(TAGE, "BTBTAGE constructor\n");
+    this->needMoreHistories = p.needMoreHistories;
```
eastonman, line 34: Why use this instead of syntax above?

eastonman (changes requested):

jensen-yan (approved):

src/cpu/pred/btb/btb_tage.cc
boran-xu: I also considered putting it in the member initialization list. But it turns out needMoreHistories is the base class  (TimeBasedPredictor) 's member, and can only be initialized in the constructor body

eastonman (approved):
LGTM

PR #398, opened 2025-05-22
cpu: fix ideal-kmhv3 config

jensen-yan:
Change-Id: I8ccee31a3b412ed392a984ca3577e034b170445f

first change icache size from 128 to 64kb


jensen-yan:
this PR have 8 commits, the ideal v3 performance impact is      
icache size  -0.05
dcache size -0.03
tage size  -0.16
new scheduler -0.37

other 4commits:
18.71-18.63=0.08
18.63-18.61=0.02
18.61-18.47=0.14
no change

new scheduler have serious impact, maybe we need revert it.

PR #402, opened 2025-05-26
cpu-o3: add constant folding opt

eastonman:
Implemented a addi elimination according to
paper RENO: A Rename-Based Instruction Optimizer
Only RENO-CF is implemented.

Change-Id: Ic656d01132d06a158b2b6682771f502184231d1b


jensen-yan:
Why can't displacement be placed in the PhysRegId class so that a new class can't be redefined and all interfaces need to be modified? This code is less intrusive.

eastonman:
Displacement is not bind to a physical register ID. Multiple instruction may share a same physical register with different displacement. So this varible need to be in rename result.

jensen-yan (approved):

PR #407, opened 2025-05-27
Fetch refactor 

jensen-yan:
working in progress, for ci tests


tastynoob (dismissed):

jensen-yan:
The previous commits were just simple functional reconstructions of fetch, and only one commit [cpu-o3: refactor Decoder and Fetch](https://github.com/OpenXiangShan/GEM5/pull/407/commits/5e985186cd0b26fcdcde0bb21b50e62ab44ca680) reconstructed decode to fix the ICache access bug, resulting in performance improvements.

image


The original logic is, when 64B aligned pc is not equal to the current index pc, then send continuous icache requests
The problem is that in the BTB architecture, since most ftq end addr are 32B aligned, most of the fetch pc are 64 aligned/32B aligned, resulting in pipelineFetch often does not send icache requests, that is, when icache hits continuously, it will not be able to access icache in a pipeline, resulting in one more beat of bubbles.
Because GEM5 here default access to ICache will have 1 beat delay, if not pipelined, will first access icache, 1 beat delay to obtain data, and then to decide the next icache access request!
The new version is changed to immediately send a pipelined icache fetch request as long as the fetch request is not in the existing fetchBuffer(icache return data), thus accessing ICache pipelined.

tastynoob (approved):

boran-xu (approved):
made fetch logic much more readable!

PR #410, opened 2025-05-28
misc: using gem5.fast mode to run CI performance test.

jensen-yan:
- Changed build target from 'gem5.opt' to 'gem5.fast' in multiple scripts to enable faster builds.
- Introduced an environment variable 'GEM5_BUILD_TYPE' to allow users to specify the build type, defaulting to 'opt' for backward compatibility.
- use GEM5_BUILD_TYPE=fast bash util/xs_scripts/kmh_v3_ideal.sh xxx to use fast mode.
- fast mode runs 20% faster than opt mode.

Change-Id: Ibe1530ec95de5f816ba3f42d51bb771b6c78d5cb

### DIFF
--- a/.github/workflows/gem5-perf-template.yml
+++ b/.github/workflows/gem5-perf-template.yml
@@ -61,7 +61,7 @@ jobs:
           cmake ..
           make -j 48
           cd $GEM5_HOME
-      - name: Build GEM5 opt
+      - name: Build GEM5 fast
         run: |
           # use pgo profile to build gem5
           export GEM5_HOME=$(pwd)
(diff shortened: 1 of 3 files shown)


jensen-yan:
https:www.gem5.org/documentation/learning_gem5/part1/building/

Built with all optimizations on (including link-time optimizations on supported platforms) and with no debug symbols. Additionally, any asserts are removed, but panics and fatals are still included. fast is the highest performing binary, and is much smaller than opt. However, fast is only appropriate when you feel that it is unlikely your code has major bugs.

fast mode will delete all asserts, I'm also worrying this might hide some bugs.

eastonman (approved):
Looks good

tastynoob (approved):

PR #416, opened 2025-06-03
cpu-pred: speed up data copy and repeat logic

tastynoob:


eastonman:
Will run some speed test and profiling on my environment.

eastonman (approved):
At least no performance degradation in my test env. LGTM.

tastynoob:
In fact, the speed is no improvement, looks like compiler can cover this pattern

jensen-yan (approved):
LGTM

PR #418, opened 2025-06-04
cpu: avoid using bitset in foldedhist

eastonman:
foldedhist is frequently allocated and deallocated, using boost::dynamic_bitset is expensive here

Also fix a bug in SC path history folding.

This PR alone reduce coremark simulation time from 30s to 24s

Baseline flame graph:

Screenshot 2025-06-04 at 17 49 01

This PR

Screenshot 2025-06-04 at 17 41 43

Conbimed with #416, coremark simluation time can be further reduced to 21s on 7950X3D machine, which is on-par with FTB KMHV2 version.

Flame graph combined with #416 

Screenshot 2025-06-04 at 17 46 29

### DIFF
--- a/src/cpu/pred/btb/folded_hist.cc
+++ b/src/cpu/pred/btb/folded_hist.cc
@@ -1,10 +1,13 @@
 #include "cpu/pred/btb/folded_hist.hh"
 
-namespace gem5 {
+namespace gem5
+{
 
-namespace branch_prediction {
+namespace branch_prediction
+{
 
-namespace btb_pred {
+namespace btb_pred
+{
 
 /**
  * Update the folded history with a new branch outcome.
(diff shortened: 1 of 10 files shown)


tastynoob:
You did what I wanted to do，nice work

tastynoob:
And also want to optimize the tabel lookup logic, now the lookup will traverse the all table to match tag

tastynoob (dismissed):

eastonman:
Multiple std::map usage is slowing down frontend for quite a bit. Maybe change that to static arrays in the near future.

tastynoob:
And now predictor fill data in stage_x to stage_end, it will call lookup multi times but each same result, we can only lookup 
table once.

jensen-yan (dismissed):
LGTM

src/cpu/pred/btb/folded_hist.cc
```diff
@@ -15,152 +18,135 @@ namespace btb_pred {
...
+                temp &= foldedMask;
 
+                // Second shift for bit 0
                 // Step 1: Handle the bits that would be lost in shift
-                temp.resize(foldedLen + 1);
-                for (int i = 0; i < 1; i++) {
-                    // XOR the highest bits from GHR with corresponding positions in folded history
-                    temp[posHighestBitsInOldFoldedHist[i]] ^= ghr[posHighestBitsInGhr[i]];
-                }
+                temp ^= (ghr[posHighestBitsInGhr[1]] << posHighestBitsInOldFoldedHist[1]);
```
boran-xu: I agree with your  using` ghr[posHighestBitsInGhr[1]] `here instead of` ghr[posHighestBitsInGhr[0]]`, as was the case before the fix. However, I think the amount to shift should still be` posHighestBitsInOldFoldedHist[0]`, since temp has already been shifted left by one bit, the bit you'd want to xor should be still at `posHighestBitsInOldFoldedHist[0]`

boran-xu: In any case, I'll rebase my phr PR onto this PR, and fix bugs if any were to be found. (I'll check the correctness of folded phr with `tage->checkFoldedHistory`) Though I expect everything to go smoothly after this line is fixed!

eastonman: Ok, will fix that

jensen-yan (approved):

boran-xu (approved):

PR #420, opened 2025-06-05
arch-riscv: add functional tlb support

happy-lx:
use --functional-tlb arg to enable it

Change-Id: Ifb84a8d53c77669c9ba589d8a988f52df2a23d4e


happy-lx:
**For now, I set --functional-tlb as True by default to Test functionalTLB. 
This should be functionally correct and useful for performance.**

Later I will set it as False

jueshiwenli (approved):

jensen-yan (approved):

PR #421, opened 2025-06-05
mem-cache: bop can parallel with stream

tastynoob:
Change-Id: I68b110e6d20e2444471194aa66e39fdede5d4018


tastynoob:
I found the bop and stream will be some mutual exclusion, and unstable. The order of cpu memory access will influence the competition betweeen bop and stream

jueshiwenli (approved):

happy-lx (approved):

PR #440, opened 2025-06-12
cpu-o3: add bypass BPU fetch after squash

jensen-yan:
- Introduced a new parameter `enableBypassBpuFetch`
- Updated the `DecoupledBPUWithBTB` class to handle the new bypass logic in the tick and enqueue processes.
- Adjusted the handling of override bubbles to accommodate the new feature.


jensen-yan:
for CI test

jensen-yan:
for ci test, no need now. New btb use state machine to enable bypass bpu.

PR #462, opened 2025-06-24
cpu-o3: Fix missed detection of split requests.

Ergou-ren:
In the RVV, the request needs to be split related processing, but the detected request also needs.
When the detected object is a split request, add corresponding processing logic.

Change-Id: I2d48b6d83d0d64b97016d60d8352175835cae89f

### DIFF
--- a/src/cpu/o3/lsq_unit.cc
+++ b/src/cpu/o3/lsq_unit.cc
@@ -79,6 +79,92 @@ namespace gem5
 namespace o3
 {
 
+LSQUnit::AddrRangeCoverage
+LSQUnit::checkStoreLoadForwardingRange(typename StoreQueue::iterator store_it,
+                                      LSQRequest *request, const DynInstPtr &load_inst,
+                                      int load_req_idx, int store_req_idx)
+{
+    // Extract address ranges from parameters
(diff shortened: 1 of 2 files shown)


tqxu23 (dismissed):

src/cpu/o3/lsq_unit.cc
```diff
@@ -79,6 +79,91 @@ namespace gem5
...
+ * @param load_idx Index of load sub-request (-1 if not split)
+ * @param store_sub_idx Index of store sub-request (-1 if not split)
+ * @param store_it Iterator to store queue entry
+ * @param request Load request pointer
+ * @param load_inst Load instruction pointer
+ * @param store_is_split Whether the store is split
+ * @return Coverage type for this address range comparison
+ */
+static LSQUnit::AddrRangeCoverage
+checkStoreLoadForwardingRange(Addr req_s, Addr req_e, Addr st_s, Addr st_e,
```
tastynoob: Should be member function

Ergou-ren: ok

tastynoob (approved):

tqxu23 (approved):

PR #476, opened 2025-07-03
arch-riscv: SV48 PTW

jueshiwenli:
Add SV48 single-level PageTableWalker support. Currently, SV48 is only supported through PTW, and the feature is disabled by default and can be enabled via an option.

Change-Id: Ie300e3bd7427c750353cec3729008c8c6ca4126b


configs/common/Options.py
```diff
@@ -290,6 +290,9 @@ def addCommonOptions(parser, configure_xiangshan=False):
     parser.add_argument("--functional-tlb", action="store_true", default=False,
                         help="""
                         Use functional TLB""")
+    parser.add_argument("--open-sv48", action="store_true", default=False,
```
tastynoob, line 293: Why 2 options here

jueshiwenli: oh let me have a review!

happy-lx (dismissed):

jensen-yan (approved):

happy-lx (approved):

tastynoob (approved):

PR #500, opened 2025-07-31
cpu: fix LSQRequest num control bugs by moving into LSQUnit

zephyrols:
During multi-core operation, we observed a situation where the number of `numSBufferRequests` exceeded the upper limit of `port->sbufferEntries`. This is because `numSBufferRequest` was wrongly defined as a `static` variable, which resulted in different lsqUnits sharing the same numSBufferRequest. As a result, the numSBufferRequest in each lsqUnit did not exceed its respective sbufferEntries, but the total exceeded the limit. Therefore, it is modified to a `non-static` variable and `moved inside the LSQUnit`. Each lsqUnit maintains its own set of values, and numSingleRequest and numSplitRequest follow the same pattern.

Change-Id: I42e41a69333ce85d3145cae13441eda0ab3cc158

### DIFF
--- a/src/cpu/o3/lsq.cc
+++ b/src/cpu/o3/lsq.cc
@@ -83,9 +83,6 @@ LSQ::DcachePort::DcachePort(LSQ *_lsq, CPU *_cpu) :
     RequestPort(_cpu->name() + ".dcache_port", _cpu), lsq(_lsq), cpu(_cpu)
 {}
 
-uint64_t LSQ::LSQRequest::numSBufferRequest = 0;
-uint64_t LSQ::LSQRequest::numSingleRequest = 0;
-uint64_t LSQ::LSQRequest::numSplitRequest = 0;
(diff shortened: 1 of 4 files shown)


tastynoob (approved):

happy-lx (approved):

zephyrols:
During multi-core operation, we observed a situation where the number of `numSBufferRequests` exceeded the upper limit of `port->sbufferEntries`. This is because `numSBufferRequest` was wrongly defined as a `static` variable, which resulted in different lsqUnits sharing the same numSBufferRequest. As a result, the numSBufferRequest in each lsqUnit did not exceed its respective sbufferEntries, but the total exceeded the limit. Therefore, it is modified to a `non-static` variable and moved inside the lsqUnit. Each lsqUnit maintains its own set of values, and numSingleRequest and numSplitRequest follow the same pattern.

PR #506, opened 2025-08-06
misc: pin scons python package version to protect compatibility

Yakkhini:
Change-Id: Ic6fcc7eca217c9d15b0fd3dccf0fd9d2814c962

Currently the newest SCons in nixpkgs has a built-in Python 3.13 library which cause error when building GEM5. Override the input python packages and pin to 3.10 can improve project robust.

### DIFF
--- a/nix/flake.nix
+++ b/nix/flake.nix
@@ -43,7 +43,7 @@
           pythonPackages.distutils
           pythonPackages.pydot
           pythonPackages.six
-          scons
+          (scons.override {python3Packages = pkgs.python310Packages;})
           sqlite
           swig
           zlib


jensen-yan (approved):

eastonman:
Did you meet any error compiling in nix environment?
Because it seems there is a similar override here https://github.com/OpenXiangShan/GEM5/blob/89bdaf785213e5329dd905548bcfd069d8eab025/nix/flake.nix#L21

Yakkhini:
> Did you meet any error compiling in nix environment? Because it seems there is a similar override here
> 
> https://github.com/OpenXiangShan/GEM5/blob/89bdaf785213e5329dd905548bcfd069d8eab025/nix/flake.nix#L21

Sorry, that's my fault. I override it in another place and didn't test in this flake. It looks like no need to override it twice.

PR #513, opened 2025-08-14
cpu-o3: sbuffer write bank accurately

tastynoob:
Change-Id: Idea3fc24cc2745e5bc2229bbd4e902a7d6f53c02

### DIFF
--- a/configs/example/xiangshan.py
+++ b/configs/example/xiangshan.py
@@ -360,6 +360,7 @@ def setKmhV3IdealParams(args, system):
         cpu.scheduler = KMHV3Scheduler()
 
         cpu.BankConflictCheck = True   # real bank conflict 0.2 score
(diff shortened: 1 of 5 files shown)


jensen-yan (dismissed):

jueshiwenli (dismissed):

happy-lx:
Would it be better to add a parameter to control whether to use accurate store bank conflict detection? This is an option on the RTL and is turned off by default due to timing reasons.

tastynoob:
> Would it be better to add a parameter to control whether to use accurate store bank conflict detection? This is an option on the RTL and is turned off by default due to timing reasons.

ok

jensen-yan (approved):

jueshiwenli (approved):

PR #559, opened 2025-09-24
fix tage

jensen-yan:
1. Fixed tage phr folding history calculation bug, there will be a problem when histLen is greater than folding history, it should only be calculated with hash of histLen length, not full hash, so as to pass history check, no performance impact.
2. Add new DebugFlags=TAGEHistory
3. Fix unit tests Adapt new PHR algorithm, run through all unit tests
4. Simplify tage two-way replacement algorithm, delete LFSR, directly query the useful bits of the two ways on the spot, and find one that is 0 and you can allocate it. Remove LFSR performance improvement of 0.03 points.
5. Add use_alt_on_na, performance improves by 0.06 points.


jensen-yan:
@codex review

src/cpu/pred/btb/btb_tage.hh
```diff
@@ -250,8 +244,11 @@ class BTBTAGE : public TimedBaseBTBPredictor
     const unsigned baseTableSize;  // Base table size
     const unsigned maxBranchPositions;  // Maximum branch positions per 64-byte block
 
-    // Table for tracking when to use alternative prediction
-    std::vector<std::vector<short>> useAlt;
+    // Table for tracking when to use alternative prediction on provider weak
+    // use_alt_on_na: indexed by PC, 7-bit signed saturating counter [-64, 63]
+    static constexpr unsigned useAltOnNaSize = 128;
+    static constexpr unsigned useAltOnNaWidth = 7;
```
eastonman: Why not add to Python param?

src/cpu/pred/btb/btb_tage.hh
```diff
@@ -260,7 +257,7 @@ class BTBTAGE : public TimedBaseBTBPredictor
     void setTag(Addr &dest, Addr src, int table);
 
     // Debug flag
-    bool debugFlagOn{false};
+    bool debugFlagOn{true};
```
eastonman: Is this needed?

jensen-yan: Required, open by default, and then use debug flags such as TAGEHisotry to control printf

eastonman: Why not just remove this flag and use gem5's debug flags only.

jensen-yan: ok, i will fix it

eastonman (approved):

Yakkhini (approved):

PR #560, opened 2025-09-24
cpu-o3: Match on taken pc/target; drop uBTB dummy and numNTConds

jensen-yan:
Rationale:
- Align stage matching with control-flow equivalence: if both predict taken at the same control PC and to the same target, treat them as equal.
- finalPred drives all history updates (GHR/PHR/Bw/Local), so early-stage shamt alignment is unnecessary for match.
- Remove the uBTB dummy mechanism and the numNTConds field used to emulate not-taken conditionals.

Changes:
- FullBTBPrediction::match: remove END and HIST_INFO comparisons; keep only taken/not-taken, controlAddr, and target checks.
- uBTB: remove insertion of dummy conditional entries and the numNTConds field/storage.
- Clean out dead code related to dummy handling.

Impact:
- Reduces overrides caused by history-length mismatches while preserving redirect equivalence.
- Unblocks placing TAGE at S0 to correct uBTB/ABTB direction without dummy artifacts.
- Avoids potential assertions or mis-indexing caused by out-of-block dummy entries.

Change-Id: I0bd7cece129aff436dea5537369663336e8ca72f

### DIFF
--- a/src/cpu/pred/btb/btb_ubtb.cc
+++ b/src/cpu/pred/btb/btb_ubtb.cc
@@ -116,16 +116,6 @@ UBTB::fillStagePredictions(const TickedUBTBEntry &entry, std::vector<FullBTBPred
     }
 
     if (entry.valid) {
-        // push back dummy conditional branches before the taken branch, to create the correct speculative history
-        // information, these dummy entries are not taken, thanks to them not being in stagePreds.condTakens.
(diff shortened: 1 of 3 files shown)


jensen-yan:
@codex review

CJ362ff (approved):

eastonman (approved):

PR #581, opened 2025-10-23
Simple mbtb perf

jensen-yan:
[[align] only update mBTB if's squash entry](https://github.com/OpenXiangShan/GEM5/commit/fa08d783a3a7de9732dd9328b99285b42881ec8f)


eastonman (approved):

Yakkhini (approved):
There is tiny performance decrease (about 0.013) when compared to mainline, but it's fine.

PR #593, opened 2025-11-05
Train MainBTB & TAGE predictor at resolve stage

Yakkhini:


Yakkhini:
> ## 🚀 Coremark Smoke Test Results
> Branch 	IPC 	Change
> Base (`xs-dev`) 	`2.1375` 	-
> This PR 	`2.0856` 	📉 `-0.0519` (`-2.43%`)
> 
> ✅ Difftest smoke test passed!

According to local coremark running and analyze, MKPI is reduced when only train up to single resolve queue entry per cycle, but there is more performance decrease in non-frontend bound.

src/cpu/pred/btb/btb_tage.cc
```diff
@@ -75,6 +76,7 @@ useAltOnNaWidth(p.useAltOnNaWidth),
 numTablesToAlloc(p.numTablesToAlloc),
 enableSC(p.enableSC),
 updateOnRead(p.updateOnRead),
+resolvedUpdate(false),
```
jensen-yan: why not use p.resolvedUpdate, set in BranchPredictor.py

src/cpu/o3/fetch.cc
```diff
@@ -1503,6 +1528,7 @@ Fetch::handleCommitSignals(ThreadID tid)
                 dbpftb->update(fromCommit->commitInfo[tid].doneFsqId, tid);
             } else if (isBTBPred()) {
                 assert(dbpbtb);
+                dbpbtb->commitStream(fromCommit->commitInfo[tid].doneFsqId);
                 dbpbtb->update(fromCommit->commitInfo[tid].doneFsqId, tid);
```
jensen-yan, line 1531: use commitStream instead of update, bpu.update is not used,
Either delete it, or let's not rename it for now, just add some comments to the update function.
This doesn't feel very elegant.

src/cpu/o3/iew.cc
```diff
@@ -1656,6 +1673,10 @@ IEW::executeInsts()
     // @todo This doesn't actually work anymore, we should fix it.
 //    printAvailableInsts();
 
+    // Clear resolvedFSQId and resolvedInstPC since they are already handled in frontend
+    ThreadID tid = *activeThreads->begin();
+    toFetch->iewInfo[tid].resolveQueue.clear();
```
jensen-yan, line 1678: The `resolveQueue` is currently a member of the forward pipeline, meaning a new queue is created for each iteration, and all processes in that iteration are completed. This is actually slightly different from the persistent queue implementation in RTL.

jensen-yan (approved):

src/cpu/o3/iew.cc
Yakkhini: Will align to RTL in future

src/cpu/pred/btb/decoupled_bpred.cc
```diff
@@ -1170,7 +1238,9 @@ DecoupledBPUWithBTB::updatePredictorComponents(FetchStream &stream)
 
         // Update all predictor components
         for (int i = 0; i < numComponents; ++i) {
-            components[i]->update(stream);
+            if (components[i] != tage || components[i] != mbtb) {
```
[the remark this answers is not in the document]
Yakkhini: That's wrong and the code is refactored now.

PR #645, opened 2025-12-11
mem: fix wrong raw violation

happy-lx:
When load is at s0 and store is at s1, wrong raw violation will occur. However, load can forward data from store when it goes into s1, so no need to do raw violation.

Change-Id: I2c68663288b2009e722516589dd34b9743e49bd6

## Summary by CodeRabbit

* **New Features**
  * Loads can conditionally skip raw-dependence checks in early pipeline stages to reduce false RAW violations.
  * Added per-stage metric tracking for skipped RAW checks to improve memory-access analysis and reporting.
  * Pipeline logic now clears the skip condition as instructions advance, improving simulation accuracy.


src/cpu/o3/lsq_unit.cc
```diff
@@ -1133,6 +1158,13 @@ LSQUnit::checkViolations(typename LoadQueue::iterator& loadIt,
                             break;
                         }
 
+                        // If load is at stage 0 while store is at stage 1, no RAW violation
+                        // should occur since the load can forward data from store later
+                        if (ld_inst->pipelineStage == 0) {
```
tastynoob: why should not use ld_inst->isTranslationComplete() instead of pipelineStage?

tastynoob (approved):

PR #666, opened 2025-12-19
cpu-o3: add more performance counter in resolve queue

Yakkhini:
## Summary by CodeRabbit

* **New Features**
  * Added new runtime metrics tracking resolve-queue enqueue events, dequeue events, and per-cycle queue occupancy to improve observability.
* **Chores**
  * Removed the older "queue full" metric and replaced it with the more granular enqueue/dequeue and occupancy measurements.

---

image

### DIFF
--- a/src/cpu/o3/fetch.cc
+++ b/src/cpu/o3/fetch.cc
@@ -251,12 +251,16 @@ Fetch::FetchStatGroup::FetchStatGroup(CPU *cpu, Fetch *fetch)
                     statistics::units::Count, statistics::units::Cycle>::get(),
              "Frontend Bandwidth Bound",
              frontendBound - frontendLatencyBound),
-    ADD_STAT(resolveQueueFullCycles, statistics::units::Count::get(),
-             "Number of cycles the resolve queue is full"),
     ADD_STAT(resolveQueueFullEvents, statistics::units::Count::get(),
              "Number of events the resolve queue becomes full"),
     ADD_STAT(resolveEnqueueFailEvent, statistics::units::Count::get(),
-             "Number of times an entry could not be enqueued to the resolve queue")
+             "Number of times an entry could not be enqueued to the resolve queue"),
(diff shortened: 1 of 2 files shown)


Yakkhini:
Master:

* 0.3c: 16.85
* 0.8c: 19.91

12 + 4 entry size:

* 0.3c: 16.88
* 0.8c: 19.91

Yakkhini:
Master:

* 0.3c: 16.85
* 0.8c: 19.91

(4 + 4) entry size:

* 0.3c: 16.85
* 0.8c: 16.91

Yakkhini:
Master:

* 0.3c: 16.85
* 0.8c: 19.91(3)

(28 + 4) entry size:

* 0.3c: 16.87
* 0.8c: 16.91(4)

Yakkhini:
Conclusion：Resolve Queue Size has no significant effect on performance. Maybe a 8 entry queue even less than this size structure should be considered, to shrink microarch area.

src/cpu/o3/fetch.cc
```diff
@@ -326,6 +330,10 @@ Fetch::FetchStatGroup::FetchStatGroup(CPU *cpu, Fetch *fetch)
             .flags(statistics::total);
         frontendBandwidthBound
             .flags(statistics::total);
+        resolveEnqueueCount
+            .init(1, 8, 1);
+        resolveQueueOccupancy
+            .init(0, 32, 1);
```
jensen-yan, line 336: It's better to use resolveQueueSize instead of use 32, but now it's ok.

jensen-yan (approved):

src/cpu/o3/fetch.cc
Yakkhini: Only static data is accepted in stats construction function. Maybe `resolveQueueSize` can be pass into function as a argument, just like what `TageStats` does. But currently the `FetchStatGroup` function has no redundant arguments, so I just use a static number `32` for now.

image

PR #671, opened 2025-12-23
ITTAGE Bank Conflict Implementation

Yakkhini:
## Summary by CodeRabbit

* **Refactor**
  * Improved branch-prediction update flow: added bank-aware tracking and a gating check that makes updates more selective, avoiding redundant updates and refining when prediction state is advanced.

### DIFF
--- a/src/cpu/pred/btb/btb_ittage.cc
+++ b/src/cpu/pred/btb/btb_ittage.cc
@@ -210,13 +210,19 @@ BTBITTAGE::getPredictionMeta() {
 bool
 BTBITTAGE::canResolveUpdate(const FetchStream &stream)
 {
+    return true;
+}
+
+void
+BTBITTAGE::doResolveUpdate(const FetchStream &stream)
+{
 
     if (predBankValid) {
         predBankValid = false;
-        return false;
+        return;
     }
 
-    return true;
+    update(stream);
 }
 
 void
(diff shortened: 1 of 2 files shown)


Yakkhini:
* Master: 16.87
* ITTAGE Blocked Bank Conflict Impl: [16.50](https://github.com/OpenXiangShan/GEM5/actions/runs/20457153395?pr=671)
* ITTAGE Dump Update Bank Conflict Impl: [16.82](https://github.com/OpenXiangShan/GEM5/actions/runs/20458770208?pr=671)

The Observation in Coremark and Microbench test scale:

ITTAEG is not work in micro frontend tests, but can be trained in larger program. I suggest in larger program there are more prediction bubbles so that ITTAGE has more chance to update itself. In another words, updates in ITTAGE become a sparse behavior. It also can be explained that why there is only about 0.05 score per GHz decrease in dump version. In 20M huge client program, ITTAGE could also be trained into a stable state, even the training time has been extremely extended.

In blocked version, the ITTAGE works because updates in resolve queue would like to wait ITTAGE ready so that they can be write into ITTAGE tables. But ITTAGE's block behavior make a negative effect on global resolve update mechanism since resolve dequeue performance is heavily decreased.

In dump version, BPU prediction & update issue is fixed since resolve dequeue is not be blocked anymore. But the ITTAGE almost not working.

PR #672, opened 2025-12-24
Using mbtb basetable align

CJ362ff:
move basetable from tage to mbtb and change mbtb entry Replacement algorithm

## Release Notes

* **New Features**
  * Added configurable branch prediction base table usage, enabling selection between base table and counter-based prediction methods.
  * Introduced dynamic branch target buffer entry selection for optimized update strategies.

* **Improvements**
  * Enhanced branch tracking with direct branch detection support.

### DIFF
--- a/configs/example/kmhv3.py
+++ b/configs/example/kmhv3.py
@@ -108,6 +108,10 @@ def setKmhV3Params(args, system):
             cpu.branchPred.mgsc.enabled = False
             cpu.branchPred.ras.enabled = True
 
+            usingMbtbBaseEiterTage = True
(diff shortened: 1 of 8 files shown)


configs/example/kmhv3.py
```diff
@@ -108,6 +108,10 @@ def setKmhV3Params(args, system):
             cpu.branchPred.mgsc.enabled = False
             cpu.branchPred.ras.enabled = True
 
+            usingMbtbBaseEiterTage = True
```
jensen-yan: This has already been set in BranchPredictor.py, so it doesn't need to be set again here.

jensen-yan (approved):

PR #765, opened 2026-02-27
cpu-o3: Add replayed load completion counter

happy-lx:
Add a load instruction replay success rate counter to evaluate the effectiveness of replay behavior and provide a reference for RTL.

Change-Id: I78887a7eb4d90ad5b5a3a870977ff9aabd621468


happy-lx:
Detailed stats:
[replay-successful-rate.xlsx](https://github.com/user-attachments/files/25618285/replay.xlsx)

According to SPEC06：
image

PR #766, opened 2026-03-02
cpu-o3: Add load value prediction framework

happy-lx:
## Background
This PR integrates the load value prediction (LVP) framework into the current codebase. **The framework and datapath logic are adapted from PR #297 contributed by @zybzzz** . To simplify maintenance, Ideal Model–related code is removed, leaving only the core load training/prediction/recovery paths and the EStride algorithm.  

Additionally, this PR introduces IdealConstantLVP, an ideal constant-value predictor with an unbounded table, inspired by analysis of Apple M3/M4/A17 LVP behavior in the [FLOP paper](https://predictors.fail/files/FLOP.pdf) paper. This is intended to explore the upper bound and parameter space of constant-value prediction.

## Key Changes
- **LVP framework integration**:  keep the load training/prediction/recovery paths and the EStride algorithm.
- **Remove Ideal Model logic**: keep only the essential framework to reduce maintenance overhead.
- **Add IdealConstantLVP**: ideal constant predictor with infinite entries for upper-bound exploration.

## Performance evaluation results of IdealConstantLVP

Performance test results show that optimal performance is achieved when the saturation counter is set to 9 bits (predicting a certain PC load on 511 identical data results). This results in a 1.24% performance improvement on spec06Int.

image

## Summary by CodeRabbit

* **New Features**
  * Added a value‑prediction subsystem for speculative load‑value prediction with selectable predictors (stride and ideal‑constant).
  * Integrated prediction values into the CPU pipeline so predictions propagate through rename, issue, fetch, and commit; mispredictions cause recovery and dependent wakeups.
  * New configuration options and runtime statistics for prediction support, predictions, corrections, coverage, and accuracy.

### DIFF
--- a/src/cpu/valuepred/valuepred_unit.hh
+++ b/src/cpu/valuepred/valuepred_unit.hh
@@ -1,0 +1,66 @@
+#ifndef __VALUEPRED_UNIT_HH__
+#define __VALUEPRED_UNIT_HH__
+
+#include <string>
+
+#include "base/statistics.hh"
+#include "cpu/valuepred/valuepred_metadata.hh"
+#include "enums/ValuePredType.hh"
+#include "params/ValuePredictor.hh"
+#include "sim/sim_object.hh"
+#include "sim/stats.hh"
+
+namespace gem5
+{
+
+namespace valuepred
+{
+
+
+
+class VPUnit : public SimObject
+{
+  private:
+    using Params = ValuePredictorParams;
+
+  public:
+    VPUnit(const Params &params);
+
+    std::string name() const { return "valuePredict.base"; }
+
+    // Do the value predict.
+    virtual VPResult valuePredict(VPPredMetaData *predMetadata) = 0;
+
+    // In commit time, update value predictor
+    virtual void updateValuePredictor(VPUpdateMetaData *updateMetadata) = 0;
+
+    // Some value predictor need speculative update to support back-to-back predict.
+    virtual void specUpdateValuePredictor(VPSpecUpdateMetaData *specupdateMetadata) = 0;
+
+    // If predict error, squash the inflight instructions in value predictor.
+    virtual void squash(const uint64_t seq_no) = 0;
+
+    // Get the value predictor type
+    virtual ValuePredType getValuePredictorType() = 0;
+
+  public:
+    struct ValuePredUnitStats : public statistics::Group
+    {
+        ValuePredUnitStats(VPUnit *vp);
+
+        statistics::Scalar VPcorrected;
+        statistics::Scalar VPpredicted;
+        statistics::Formula VPaccuracy;
+        statistics::Scalar VPsupported;
+        statistics::Formula VPcoverage;
+
+    } stats;
+};
+
+
+}
+
+
+}
+
+#endif
--- a/configs/example/idealkmhv3.py
+++ b/configs/example/idealkmhv3.py
@@ -19,6 +19,8 @@ from common import Simulation
 from common.Caches import *
 from common.xiangshan import *
 
+from m5.objects.ValuePredictor import *
+
 def setKmhV3IdealParams(args, system):
     for cpu in system.cpu:
 
@@ -62,6 +64,9 @@ def setKmhV3IdealParams(args, system):
         cpu.BankConflictCheck = True
         cpu.sbufferBankWriteAccurately = True
         cpu.DcacheSetDivNum = 2
+
+        # value predictor
+        cpu.valuePred = IdealConstantLVP()
 
         # lsq
         cpu.LQEntries = 128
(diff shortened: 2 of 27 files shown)


zybzzz:
1. I used a factory pattern when designing the value predictor, and it looks like this part of the code is still in place. I’m not sure whether this design is reasonable—based on my own use, it often feels a bit awkward. I’m not sure how you feel about it either; it might be worth refactoring it with AI.

2. The EStride code largely references the open-source implementation from CVP1, and I made some modifications. If this part is eventually open-sourced, would it be better to explicitly mention that some implementation details are referenced from CVP1? Or should we simply drop EStride and keep only your self-designed VP part? When I originally wrote this part, I didn’t review it in much detail either.

@happy-lx

jueshiwenli:
> 1. I used a factory pattern when designing the value predictor, and it looks like this part of the code is still in place. I’m not sure whether this design is reasonable—based on my own use, it often feels a bit awkward. I’m not sure how you feel about it either; it might be worth refactoring it with AI.
> 2. The EStride code largely references the open-source implementation from CVP1, and I made some modifications. If this part is eventually open-sourced, would it be better to explicitly mention that some implementation details are referenced from CVP1? Or should we simply drop EStride and keep only your self-designed VP part? When I originally wrote this part, I didn’t review it in much detail either.
> 
> @happy-lx

Thank you for your feedback. We will address and fix these issues in the next revision.

The original motivation for merging this version was to provide a functional value prediction framework for their team, so that future optimizations and further work can be built on top of it.

We sincerely appreciate your suggestions. We will keep you updated as we make further progress.

happy-lx:
> 1. I used a factory pattern when designing the value predictor, and it looks like this part of the code is still in place. I’m not sure whether this design is reasonable—based on my own use, it often feels a bit awkward. I’m not sure how you feel about it either; it might be worth refactoring it with AI.
> 2. The EStride code largely references the open-source implementation from CVP1, and I made some modifications. If this part is eventually open-sourced, would it be better to explicitly mention that some implementation details are referenced from CVP1? Or should we simply drop EStride and keep only your self-designed VP part? When I originally wrote this part, I didn’t review it in much detail either.
> 
> @happy-lx

1. I feel the factory pattern is fine to me, It might be possible to improve it using AI later; As long as the framework works, the code style issue doesn't seem particularly urgent right now😆.
2. Thanks for the reminder! I think we can explicitly mention that in the copyright part and the comments. Do you think this is feasible?

@zybzzz Anyway, thank you very much for providing the LVP framework implementation!  it's very helpful for our exploration of the LVP algorithm.

zybzzz:
> > 1. I used a factory pattern when designing the value predictor, and it looks like this part of the code is still in place. I’m not sure whether this design is reasonable—based on my own use, it often feels a bit awkward. I’m not sure how you feel about it either; it might be worth refactoring it with AI.
> > 2. The EStride code largely references the open-source implementation from CVP1, and I made some modifications. If this part is eventually open-sourced, would it be better to explicitly mention that some implementation details are referenced from CVP1? Or should we simply drop EStride and keep only your self-designed VP part? When I originally wrote this part, I didn’t review it in much detail either.
> > 
> > @happy-lx
> 
> 1. I feel the factory pattern is fine to me, It might be possible to improve it using AI later; As long as the framework works, the code style issue doesn't seem particularly urgent right now😆.
> 2. Thanks for the reminder! I think we can explicitly mention that in the copyright part and the comments. Do you think this is feasible?
> 
> @zybzzz Anyway, thank you very much for providing the LVP framework implementation! it's very helpful for our exploration of the LVP algorithm.

Yes—adding a brief comment in the code file to explain it should be enough. I’m glad my code could be useful for reference, even though value prediction can be a little frustrating at times :) @happy-lx

jueshiwenli (dismissed):

src/cpu/valuepred/valuepred_unit.hh
```diff
@@ -0,0 +1,69 @@
...
+  public:
+    struct ValuePredUnitStats : public statistics::Group
+    {
+        ValuePredUnitStats(VPUnit *vp);
+
+        statistics::Scalar VPcorrected;
+        statistics::Scalar VPpredicted;
+        // Value prediction may result in unaligned memory accesses,
+        // and we need to record this.
+        statistics::Scalar faultLoad;
```
zybzzz: Value prediction only predicts values, and by itself it will not trigger exceptions such as misaligned accesses. If a value prediction is wrong, subsequent consumer instructions may compute an incorrect address; however, when the misprediction is detected and a pipeline flush is triggered, those consumer instructions are flushed as well. Therefore, this statistic is not meaningful—it was my mistake to consider it that way at the time.

zybzzz (dismissed):
Overall, there are no major issues. That invalid statistic in my comment can be removed.

src/cpu/valuepred/valuepred_unit.hh
happy-lx: > Value prediction only predicts values, and by itself it will not trigger exceptions such as misaligned accesses. If a value prediction is wrong, subsequent consumer instructions may compute an incorrect address; however, when the misprediction is detected and a pipeline flush is triggered, those consumer instructions are flushed as well. Therefore, this statistic is not meaningful—it was my mistake to consider it that way at the time.

Indeed, I agree with you. Thank you for your review! I will remove that part.

PR #769, opened 2026-03-04
cpu-o3: reduce ftq size to 64

Yakkhini:
Change-Id: Ida30dab747aaf11d36b38637476b1c137e589942

## Chores
* Optimized branch prediction system configuration parameters to improve resource efficiency.


Yakkhini:
Performance change: 20.36 -> 20.38 which is tiny. There is no additional error in smoke test so maybe we should align this parameter to RTL.

In 2-taken evaluation there is significant FTQ full event num increase, maybe caused by shrinking of FTQ size or improvement of BPU supply ability. The previous reason should be excluded.

jensen-yan (approved):

PR #780, opened 2026-03-10
cpu-o3: multi thread shared storebuffer

tastynoob:
Change-Id: Ic6b89d58d55f18268f17612fa09e02d9e27f0349

## Summary by CodeRabbit

* **Refactor**
  * Centralized store-buffer management with per-entry lifecycle, eviction reasons (flush, full, queue-full, timeout), capacity-aware eviction, coordinated per-cycle offload, retry, and eviction accounting.

* **Behavior**
  * Flush/writeback now routes through the centralized store-buffer flow.
  * Commit readiness relaxed to allow earlier commits for safe, non-first group instructions, reducing stalls.

### DIFF
--- a/src/cpu/o3/lsq.cc
+++ b/src/cpu/o3/lsq.cc
@@ -46,6 +46,7 @@
 #include <csignal>
 #include <cstdint>
 #include <cstdio>
+#include <cstring>
 #include <list>
 #include <string>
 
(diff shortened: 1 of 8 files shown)


src/cpu/o3/lsq.cc
```diff
@@ -88,12 +89,211 @@ LSQ::DcachePort::DcachePort(LSQ *_lsq, CPU *_cpu) :
...
+LSQ::StoreBuffer::insert(StoreBufferEntry *entry)
+{
+    int index = entry->index;
+    ThreadID tid = entry->tid;
+    Addr addr = entry->blockPaddr;
+    assert(_size < data_vec.size());
+    assert(!data_vld[index]);
+    assert(!lru_index.full());
+    _size++;
+    auto [it, _] = data_map.insert({hashKey(tid, addr), data_vec[index]});
```
jensen-yan, line 210: We cache this `unordered_map` iterator in `crossRef`, but `StoreBuffer::setData()` never reserves `data_map`. Once later inserts trigger a rehash, every iterator saved in `crossRef` becomes invalid, and `release()` eventually does `data_map.erase(crossRef[index])` with undefined behavior. Please reserve the map up front or store a stable key/pointer instead of an iterator.

src/cpu/o3/lsq_unit.cc
```diff
@@ -2142,13 +1965,13 @@ LSQUnit::offloadToStoreBuffer()
                 }
             }
             if (request->_numOutstandingPackets == request->_reqs.size()) {
+                ++accepted_entries;
```
jensen-yan, line 1968: `max_entries` is a store-buffer-entry budget, but a split store can call `insertStoreBuffer()` once per fragment before we increment `accepted_entries` here. That lets one wide split store consume multiple shared-buffer entries while only being charged once, bypassing both the per-cycle throttle and the SMT fairness arbitration.

jensen-yan (changes requested):
Thanks for the refactor. I found two blocking issues in the new shared-store-buffer path:

- `crossRef` caches `unordered_map` iterators, but `data_map` is never reserved, so a later rehash can make `release()` undefined.
- split stores are only charged once against the per-cycle entry budget even when they allocate multiple shared-buffer entries, which breaks the throttle/fairness logic.

One more thing worth fixing before merge: stores now leave the SQ before cache writeback completes, but `LSQ::isDrained()` / `takeOverFrom()` still ignore the shared sbuffer state, so drain/handoff can succeed with buffered stores still outstanding.

jensen-yan (approved):

PR #810, opened 2026-04-02
Full resolve train

Yakkhini:
Draft.

## Summary by CodeRabbit

* **New Features**
  * Full packet-based resolve-train feedback with runtime toggles, new predictor hooks, per-instruction metadata, and additional statistics.
  * Per-fetch-target generation/identity tracking to validate training against FTQ entries.

* **Refactor**
  * Dual-path resolve handling supporting legacy resolved-update and full resolve-train workflows; predictor interfaces extended to accept resolved-train packets.

* **Bug Fixes**
  * Prevent IEW signal double-processing per tick and ensure unmatched packets are cleaned up.

* **Tests**
  * New unit tests covering resolve-train behavior and bank-conflict scenarios.

* **Documentation**
  * Added a frontend full resolve-train review guide.


src/cpu/o3/fetch.cc
```diff
@@ -1454,67 +1492,192 @@ Fetch::checkSignalsAndUpdate(ThreadID tid)
 void
 Fetch::handleIEWSignals()
 {
+    if (lastIewSignalHandleTick == curTick()) {
```
Yakkhini, line 1471: Is this necessary？

Yakkhini:
It does not appear that disabling the resolve train causes any significant performance degradation. However, this PR still introduces a substantial increase in source code size, even after cleanup and simplification, mainly due to two features: packet rebuild in fetch and the split resolve train interface in the affected predictors, which no longer reuses `update()`.

I now want to run the 0.8 coverage test with the resolve train enabled to determine whether there is any performance issue in the new resolve train path.

https://github.com/OpenXiangShan/GEM5/actions/runs/24176533553

Yakkhini:
We never evaluate resolve train on ideal config before. Although there has about 0.2 score decrease appears on this resolve train implementation, but it still better than legacy one (0.3 score decrease).

Ref: https://github.com/OpenXiangShan/GEM5/pull/823#issuecomment-4220569542

Besides, there is also about 0.2 score increase on kmhv3 config (16.9 - 17.1).

Yakkhini:
SPEC Regression Performance Evaluation after MicroTAGE fix:

* idealkmhv3: https://github.com/OpenXiangShan/GEM5/actions/runs/24436494542
* kmhv3: https://github.com/OpenXiangShan/GEM5/actions/runs/24436503586

PR #828, opened 2026-04-14
[codex] cpu: Align BTB TAGE useful and allocation semantics

jensen-yan:
## What changed

This PR applies a small BTB TAGE semantic alignment pass against the current XiangShan RTL behavior.

It intentionally keeps the scope narrow:

- remove gem5-only local useful-bit clearing paths
- suppress longer-history allocation when the provider already comes from the highest table
- allow allocation to replace strong-but-not-useful victims directly
- add focused unit tests for the aligned behavior

## Why

Recent gcc15-spec06-0.3c comparisons show that the major gem5/RTL MPKI gap has already largely closed on xs-dev. This makes it a good point to upstream the remaining low-complexity, known semantic mismatches without bringing along the heavier trace and diagnostics work used during investigation.

Concretely, current xs-dev still differed from RTL in a few places:

- useful could be cleared by a local "humility" path even though RTL only increments useful locally
- useful could be cleared when the provider counter became weak, which RTL does not do locally
- allocation could age a strong not-useful entry instead of replacing it directly
- a mispredict on the highest history table could still fall through to final allocation failure accounting

## Impact

This keeps BTB TAGE behavior closer to RTL while keeping code complexity controlled:

- no useful width migration
- no reset cadence change
- no large trace or stats additions

The change is meant to reduce semantic drift, not to force a broader predictor refactor.

## Validation

Built and ran:

```bash
./build/RISCV/cpu/pred/btb/test/tage.test.debug
```

Observed:

- 25 tests from 3 test suites passed

Added coverage for:

- useful should not clear just because a strong correct alternative exists
- useful should not clear just because the provider becomes weak
- highest-table providers should suppress longer-history allocation
- strong but not-useful victims should be directly replaceable

## Summary by CodeRabbit

* **Tests**
  * Added test coverage for branch prediction "useful" bit behavior and allocation scenarios.

* **Improvements**
  * Refined branch prediction allocation logic and decision-making for improved accuracy.

### DIFF
--- a/src/cpu/pred/btb/btb_tage.cc
+++ b/src/cpu/pred/btb/btb_tage.cc
@@ -487,27 +487,12 @@ BTBTAGE::updatePredictorStateAndCheckAllocation(const BTBEntry &entry,
         // Update prediction counter
         updateCounter(actual_taken, 3, way.counter);
 
-        // Update useful bit based on several conditions
+        // Match RTL behavior: useful only increases when the provider proves
+        // itself against the alternative prediction. There is no local
+        // decrement/reset path tied to weak counters or "humility" cases.
         bool main_is_correct = main_info.taken() == actual_taken;
-        bool alt_is_correct_and_strong = alt_info.found &&
-                                     (alt_info.taken() == actual_taken) &&
-                                     (abs(2 * alt_info.entry.counter + 1) == 7);
-
-        // a. Special reset (humility mechanism)
-        if (alt_is_correct_and_strong && main_is_correct) {
-            way.useful = 0;
-            DPRINTF(TAGEUseful, "useful bit reset to 0 due to humility rule\n");
-        } else if (main_info.taken() != alt_taken) {
-            // b. Original logic to set useful bit high
-            if (main_is_correct) {
-                way.useful = 1;
-            }
-        }
-
-        // c. Reset u on counter sign flip (becomes weak)
-        if (way.counter == 0 || way.counter == -1) {
-            way.useful = 0;
-            DPRINTF(TAGEUseful, "useful bit reset to 0 due to weak counter\n");
+        if (main_info.taken() != alt_taken && main_is_correct) {
+            way.useful = 1;
         }
         DPRINTF(TAGE, "useful bit is now %d\n", way.useful);
 
@@ -553,6 +538,12 @@ BTBTAGE::updatePredictorStateAndCheckAllocation(const BTBEntry &entry,
         return false;
     }
 
+    // Match RTL: a provider from the highest history table should not trigger
+    // longer-history allocation.
+    if (main_info.found && main_info.table == numPredictors - 1) {
+        return false;
+    }
+
     // Special case: provider is weak but direction is correct
     // In this case, provider just needs more training, not a longer history table
     // This avoids wasteful allocation and prevents ping-pong effects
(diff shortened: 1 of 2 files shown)


jensen-yan:
Historical perf note for the force-pushed commit `693362bb70` (completed run `24340552235`, archive `20260413_204141_693362bb70_run480`).

Comparison target is the current `xs-dev` aligned baseline `eb6a3fb2eb` (completed run `24332606004`, archive `20260413_181611_eb6a3fb2eb_run474`).

I reprocessed both archives with the same local `gem5_data_proc` flow.

## Score delta

| scope | run480 | run474 | delta |
| --- | ---: | ---: | ---: |
| overall | 19.90647 | 19.92441 | -0.01794 |
| int | 18.68605 | 18.73395 | -0.04790 |
| fp | 20.81563 | 20.81000 | +0.00563 |

This is a small regression overall (about `-0.09%`), concentrated on the integer side.

## Main benchmark drops

| benchmark | score delta |
| --- | ---: |
| gobmk | -0.23633 |
| mcf | -0.10762 |
| soplex | -0.08914 |
| omnetpp | -0.06346 |
| sjeng | -0.06036 |
| gcc | -0.02324 |
| dealII | -0.01983 |

## Main slice regressions

| slice | cycle delta | cond_MPKI delta | probe allocFailure delta | resetUseful delta |
| gobmk_13x13_9165 | +480742 | +1.27320 | +637173 | -4 |
| mcf_10366 | +319170 | +0.56040 | -20199 | -79 |
| gobmk_trevorc_10857 | +290275 | +0.75880 | +210118 | -1 |
| gobmk_nngs_453 | +279186 | +0.75510 | +389335 | -2 |
| soplex_pds-50_13064 | +182782 | +0.29715 | -34710 | +2 |
| sjeng_71271 | +57152 | +0.11965 | +15397 | 0 |
| sjeng_22213 | +53183 | +0.15725 | +26907 | -1 |
| dealII_6407 | +53781 | +0.26305 | -46098 | +1 |

Note: `allocateFailure` here is the probe-level no-eligible-victim counter, not the final no-valid-table failure.

## Why gobmk plausibly regresses

Looking at the current light-align code:

- useful now only increases and no longer has the local clear paths that xs-dev had before:
  - `src/cpu/pred/btb/btb_tage.cc:490-496`
- useful stays set until the global reset path fires:
  - `src/cpu/pred/btb/btb_tage.cc:639-649`
- replacement priority is now `invalid -> weak&&!useful -> any !useful`:
  - `src/cpu/pred/btb/btb_tage.cc:577-620`

So for a partial alignment like this one, the reasonable expectation is:

1. entries become stickier than on xs-dev,
2. some branch-heavy slices see more victim-search pressure before reset catches up,
3. this shows up as higher probe-level allocation pressure and worse branch MPKI.

That is exactly what the worst `gobmk` slices show. For example `gobmk_13x13_9165` has:

- `cond_MPKI +1.2732`
- `BPAllWrong +25447`
- `allocateFailure +637173`
- `recovery_bubble +2033008`

So I do think the `gobmk` regression is explainable from the code shape. It does **not** look like an obviously inconsistent or random regression.

The important nuance is that this PR is still a **partial** semantic alignment:

- it removes local useful clearing,
- but it does **not** align useful width,
- and it does **not** align reset cadence.

That means the new behavior can be stickier than the old xs-dev model while still not fully matching the current RTL aging behavior. `gobmk` seems particularly sensitive to that.

Also, not every regressing benchmark follows the exact same pattern. `mcf_10366` and `soplex_pds-50_13064` regress with higher `cond_MPKI`, but their probe-level `allocateFailure` does not increase in the same way as `gobmk`, so the regression is not explained by one single counter alone.

Current PR head `0296ef5590` only amends readability, and its new perf run `24377625146` is still in progress. I would expect the same performance signature unless that run says otherwise.

CJ362ff (approved):

PR #857, opened 2026-05-20
[codex] misc: Fix Python runtime detection

jensen-yan:
## Motivation

Some builds can fail during SCons configure with `Checking Python version... no` even after `Python.h` is found. This is not necessarily a Python version problem: the configure probe links an embedded Python test and then runs it, so non-system Python installations such as conda also need their `libpython` directory to be visible to the runtime loader.

The README and quick-start docs also still recommended installing a Python 3.8 conda environment for high Python versions, which is misleading now that Ubuntu 24.04's default Python 3.12 works.

## Approach

- Parse `python*-config` output with `shlex.split()`.
- Reuse absolute `-L` paths from `python*-config` as runtime library paths for embedded Python configure probes and resulting binaries.
- Remove the old Python 3.8/miniconda guidance from README and quick start.
- Update the documented recommended environment to Ubuntu 24.04 with system Python 3.12.

## Validation

- `python3 -m py_compile SConstruct`
- Forced SCons configure reached and passed `Checking Python version... 3.12.3`; the bounded build command later timed out during generated-header work, not during Python detection.
- The generated Python configure probe includes RUNPATH entries derived from `python3-config` `-L` paths and runs successfully, printing `3.12.3`.

## Summary by CodeRabbit

* **Documentation**
  * Updated build/setup instructions to target Ubuntu 24.04 and recommend the default system Python 3.12
  * Improved troubleshooting guidance for configuration-stage errors with clearer diagnostics
  * Removed legacy alternative Python-version setup instructions for a simpler environment

* **Improvements**
  * Better handling of embedded-Python library locations so runtime dependencies are found more reliably

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenXiangShan/GEM5/pull/857?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,22 +193,17 @@ If you have problem generating SPECCPU checkpoints, following links might help y
 
 Install dependencies as [official GEM5 tutorial](https://www.gem5.org/documentation/general_docs/building) says:
 
-### Setup on Ubuntu 22.04
-If compiling gem5 on Ubuntu 22.04, or related Linux distributions, you may install all these dependencies using APT:
+### Setup on Ubuntu 24.04
+
+Ubuntu 24.04 is the recommended environment for building and running this
+repository. The default system Python 3.12 works with the build flow; no
+separate Python environment is required.
 
 ``` shell
+sudo apt update
 sudo apt install build-essential git m4 scons zlib1g zlib1g-dev \
     libprotobuf-dev protobuf-compiler libprotoc-dev libgoogle-perftools-dev \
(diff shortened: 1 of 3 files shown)


tastynoob (dismissed):

jensen-yan:
Addressed in `39fdfc2efa`: removed the global `ENV['LD_LIBRARY_PATH']` mutation and now rely on the `RPATH` derived from `python*-config -L` paths. This keeps the runtime path scoped to the linked Python probe/binary and avoids contaminating later SCons child processes.

PR #869, opened 2026-05-26
Global phr v2 align

CJ362ff:
## Summary by CodeRabbit

* **Refactor**
  * Improved branch prediction path-history hashing and folding to a smaller footprint and consistent bit extraction.
  * Made squash and misprediction recovery deterministic by using the actual path-taken outcome when restoring/updating histories.

* **Tests**
  * Updated validation tests and test data to match the new path-history extraction and recovery semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenXiangShan/GEM5/pull/869?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/cpu/pred/btb/common.hh
+++ b/src/cpu/pred/btb/common.hh
@@ -403,6 +403,12 @@ struct FetchTarget
         return std::make_pair(shamt, cond_taken);
     }
 
+    bool getPHistTakenDuringSquash(Addr squash_pc, bool actually_taken) const
+    {
+        auto ctrl_pc = getControlPC();
+        return actually_taken && ctrl_pc == squash_pc;
+    }
+
     // should be called before components update
     void setUpdateInstEndPC(unsigned predictWidth)
     {
--- a/src/cpu/pred/btb/decoupled_bpred.cc
+++ b/src/cpu/pred/btb/decoupled_bpred.cc
@@ -715,14 +715,11 @@ DecoupledBPUWithBTB::pHistShiftIn(int shamt, bool taken, boost::dynamic_bitset<>
     if (shamt == 0) {
         return;
     }
-    if(taken){
-        // Calculate path hash
-        uint64_t hash = pathHash(pc, target);
-
+    if (taken) {
+        const uint64_t footprint = pathHash(pc, target);
         history <<= shamt;
         for (auto i = 0; i < pathHashLength && i < history.size(); i++) {
-            history[i] = (hash & 1) ^ history[i];
-            hash >>= 1;
+            history[i] = ((footprint >> i) & 1) ^ history[i];
         }
     }
 }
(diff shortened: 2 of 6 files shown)


CaoJiaming776:
This PHR change is based on a public reverse-engineered model of the Neoverse V2 path-history register (PHR), rather than an Arm-official microarchitectural specification.

Reference:
- Jiajie Chen, “Reverse engineered conditional branch predictors”
  https://jia.je/cpu/cbp.html

Relevant PHR data used here:
- Neoverse V2 PHR length: 528 bits
- Public reverse-engineered update rule:
  `PHR_new = (PHR_old << 2) xor footprint`
- The published reverse-engineered footprint uses branch-PC bits and target-PC bits together, rather than a simple taken/not-taken token.

Important caveat:
- This is a public reverse-engineering result, not Arm RTL or an Arm architectural guarantee.
- The intent of this patch is to make our shared BTB-side path-history update semantics closer to the best public Neoverse V2 PHR model currently available.

CaoJiaming776:
We evaluated two intermediate PHR footprint variants against the original V2-like 10-bit baseline.

Motivation
- RISC-V supports the compressed instruction set (RVC), so control-flow addresses are naturally 2-byte aligned rather than strictly 4-byte aligned.
- Because of this, we wanted to test whether including one additional low-order address bit in the PHR footprint could improve path discrimination.
- In other words, the experiments below were intended to check whether shifting the branch-PC / target-PC bit windows downward by one bit, or slightly widening the footprint, could better capture compressed-instruction control-flow patterns.

Reference baseline
- Footprint width: 10 bits
- Bit range: `B[17:2]` and `T[11:2]`
- Update style: taken-only, `PHR <<= 2`, then XOR footprint
- This is the version kept as the final baseline because it is the closest to the public reverse-engineered Neoverse V2 PHR model.

Attempt 1
- Commit: `272fcd847a`
- Footprint width: 10 bits
- Bit range: `B[16:1]` and `T[10:1]`
- Difference vs. baseline:
  - Same 10-bit width
  - Same sparse-XOR style
  - But the branch-PC and target-PC bit windows were both shifted down by 1 bit, so one additional low-order aligned bit was included
- Result on SPEC CPU2006 0.3c int:
  - about `-0.02` compared with the original 10-bit V2-like baseline

Attempt 2
- Commit: `97cb4c3311`
- Footprint width: 11 bits
- Bit range: `B[17:1]` and `T[11:1]`
- Difference vs. baseline:
  - Expanded from 10 bits to 11 bits
  - Broadened the source bit range by including one additional low bit on both branch PC and target
- Result on SPEC CPU2006 0.3c int:
  - slightly positive compared with the original 10-bit V2-like baseline

Conclusion
- The experiments were motivated by the presence of 2-byte-aligned compressed instructions in RISC-V.
- Shifting the original 10-bit footprint window downward by one bit caused a small regression.
- Expanding the footprint to 11 bits showed a slight improvement, but we still reverted to the original 10-bit V2-like footprint in order to stay closest to the public reverse-engineered Neoverse V2 model.

Restore commit
- Commit: `e9d029c687`
- This commit restores the shared BTB PHR footprint back to the original 10-bit V2-like version.

jensen-yan:
看最新的实现，事实上本PR 的0.1分提升来自于一个bug 的修复，新版PHR 算法已经在RTL 上实现但没有收益（+0.02 /GHz),  在模型上也没有收益，PHR 算法或许还有探索空间等待未来继续。

PR #898, opened 2026-06-04
configs: Centralize prefetcher configuration policy

XXtaoo:
## Summary
- Move XSComposite and L2Composite component enable policy into PrefetcherConfig helpers while keeping SimObject defaults off.
- Centralize the pf-buffer default in PrefetcherConfig and add explicit enable/disable CLI overrides.
- Remove duplicated pf-buffer assignments from kmhv2, kmhv3, and idealkmhv3.

## Validation
- python3 -B -m compileall -q configs/common/PrefetcherConfig.py configs/common/Options.py configs/example/kmhv2.py configs/example/kmhv3.py configs/example/idealkmhv3.py src/mem/cache/prefetch/Prefetcher.py
- git diff --check -- configs/common/PrefetcherConfig.py configs/common/Options.py configs/example/kmhv2.py configs/example/kmhv3.py configs/example/idealkmhv3.py src/mem/cache/prefetch/Prefetcher.py
- Lightweight stub checks for pf-buffer tri-state behavior and XSComposite/L2Composite profile switches.

Full gem5 build or checkpoint simulation was not run.

## Summary by CodeRabbit

* **New Features**
  * CLI now enables prefetch buffers by default and adds an explicit flag to disable them.

* **Changes**
  * Example configs updated to respect the new CLI default (no longer force-enable buffers).
  * Prefetcher behavior made more conservative: several prefetch features disabled by default, composite prefetch tuning simplified, and queue/request limits adjusted to reduce aggressive prefetching.

### DIFF
--- a/configs/common/Options.py
+++ b/configs/common/Options.py
@@ -306,9 +306,16 @@ def addCommonOptions(parser, configure_xiangshan=False):
                         help="""
                         Prefetching cache level for SMS'pht""")
 
-    parser.add_argument("--enable-pf-buffer", action="store_true", default=False,
+    parser.set_defaults(enable_pf_buffer=None)
+    parser.add_argument("--enable-pf-buffer", action="store_true",
+                        dest="enable_pf_buffer",
                         help="""
                         Force all hardware prefetchers to enable their
+                        optional prefetch buffer (QueuedPrefetcher.use_pf_buffer).""")
+    parser.add_argument("--disable-pf-buffer", action="store_false",
+                        dest="enable_pf_buffer",
+                        help="""
+                        Force all hardware prefetchers to disable their
                         optional prefetch buffer (QueuedPrefetcher.use_pf_buffer).""")
 
     parser.add_argument("--cpu-clock", action="store", type=str,
--- a/configs/common/Options.py
+++ b/configs/common/Options.py
@@ -306,14 +306,8 @@ def addCommonOptions(parser, configure_xiangshan=False):
                         help="""
                         Prefetching cache level for SMS'pht""")
 
-    parser.set_defaults(enable_pf_buffer=None)
-    parser.add_argument("--enable-pf-buffer", action="store_true",
-                        dest="enable_pf_buffer",
-                        help="""
-                        Force all hardware prefetchers to enable their
-                        optional prefetch buffer (QueuedPrefetcher.use_pf_buffer).""")
     parser.add_argument("--disable-pf-buffer", action="store_false",
-                        dest="enable_pf_buffer",
+                        dest="enable_pf_buffer", default=True,
                         help="""
                         Force all hardware prefetchers to disable their
                         optional prefetch buffer (QueuedPrefetcher.use_pf_buffer).""")
--- a/configs/common/PrefetcherConfig.py
+++ b/configs/common/PrefetcherConfig.py
@@ -5,118 +5,175 @@ from common import ObjectList
 from m5.objects.Prefetcher import XSPhysicalSmallBOP, XSVirtualLargeBOP
 
 
+DEFAULT_ENABLE_PF_BUFFER = True
+
+
 def _get_hwp(hwp_option):
     if hwp_option == None:
         return NULL
     hwpClass = ObjectList.hwp_list.get(hwp_option)
     return hwpClass()
 
+def is_pf_buffer_enabled(options):
+    option_value = getattr(options, 'enable_pf_buffer', None)
+    if option_value is None:
+        return DEFAULT_ENABLE_PF_BUFFER
+    return option_value
+
+def _configure_pf_buffer(prefetcher, pf_buffer_enabled):
+    if prefetcher != NULL and hasattr(prefetcher, 'use_pf_buffer'):
+        prefetcher.use_pf_buffer = pf_buffer_enabled
+
+def _set_pf_buffer_training_policy(prefetcher, pf_buffer_enabled):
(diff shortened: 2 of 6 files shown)


configs/common/Options.py
```diff
@@ -306,10 +306,17 @@ def addCommonOptions(parser, configure_xiangshan=False):
...
                         Prefetching cache level for SMS'pht""")
 
-    parser.add_argument("--enable-pf-buffer", action="store_true", default=False,
+    parser.set_defaults(enable_pf_buffer=None)
+    parser.add_argument("--enable-pf-buffer", action="store_true",
+                        dest="enable_pf_buffer",
                         help="""
                         Force all hardware prefetchers to enable their
                         optional prefetch buffer (QueuedPrefetcher.use_pf_buffer).""")
+    parser.add_argument("--disable-pf-buffer", action="store_false",
```
jueshiwenli, line 309: I don’t quite understand the difference between the --enable-pf-buffer and --disable-pf-buffer options. Does one enable the pf-buffer and the other disable it? If so, theoretically, would it be enough to keep only one of them?

XXtaoo: This two options are indeed confusing because  enable_pf_buffer is now tri-state:
not specified means using PrefetcherConfig's default config, while --enable-pf-buffer and --disable-pf-buffer are explicit overrides.

Ergou-ren: If it is a choice of three states, can't it be controlled directly with parameters, without using confusing control parameters like enable and disable?

configs/common/PrefetcherConfig.py
```diff
@@ -5,118 +5,175 @@
...
 def _get_hwp(hwp_option):
     if hwp_option == None:
         return NULL
     hwpClass = ObjectList.hwp_list.get(hwp_option)
     return hwpClass()
 
+def is_pf_buffer_enabled(options):
+    option_value = getattr(options, 'enable_pf_buffer', None)
+    if option_value is None:
+        return DEFAULT_ENABLE_PF_BUFFER
```
Ergou-ren: Can we directly return True here and add a comment indicating that the parameter DEFAULT_ENABLE_PF_BUFFER seems strange?

configs/common/PrefetcherConfig.py
```diff
@@ -5,118 +5,175 @@
...
+        assert prefetcher_name == 'PrefetcherForwarder'
+
+def _configure_l2_wrapper_prefetcher(prefetcher, prefetcher_name, options,
+                                     pf_buffer_enabled):
+    if not options.classic_l2:
+        _configure_l2_composite(prefetcher, prefetcher_name, options)
+        _set_pf_buffer_training_policy(prefetcher, pf_buffer_enabled)
+        if options.l1_to_l2_pf_hint:
+            prefetcher.queue_size = 32
+            prefetcher.max_prefetch_requests_with_pending_translation = 128
```
Ergou-ren, line 126: if options.classic_l2:
if not options.classic_l2:
It seems to be mutually exclusive.

configs/common/PrefetcherConfig.py
```diff
@@ -5,118 +5,175 @@
...
+        _configure_l2_wrapper_prefetcher(prefetcher, prefetcher_name, options,
+                                         pf_buffer_enabled)
 
     if cache_level == 'l3':
-        if options.l2_to_l3_pf_hint:
-            prefetcher.queue_size = 64
-            prefetcher.max_prefetch_requests_with_pending_translation = 128
+        _configure_l3_prefetcher(prefetcher, options)
 
     return prefetcher
```
Ergou-ren, line 164: In this file, many functions have been added to control the prefetch configuration.
Add some comments to specific functions to make them more understandable.

configs/common/Options.py
XXtaoo: Ok,I will remove this confusing tri-stare config, only use -disable-pf-buffer to disable pf buffer.

configs/common/PrefetcherConfig.py
XXtaoo: Ok,I will remove this confusing tri-stare config, only use -disable-pf-buffer to disable pf buffer.

configs/common/PrefetcherConfig.py
XXtaoo: OK, I will add some comments for this functions

configs/common/PrefetcherConfig.py
XXtaoo: I will modify this part of the code to make it easier to understand.

Ergou-ren (dismissed):

configs/common/PrefetcherConfig.py
```diff
@@ -6,117 +6,159 @@
...
+
+def _register_prefetcher_tlb(prefetcher, cpu):
+    if cpu != NULL:
+        prefetcher.registerTLB(cpu.mmu.dtb, cpu.mmu.functional)
+
+def _configure_xs_composite_common(prefetcher, options):
+    # Keep only option/profile-dependent overrides here. Stable model defaults
+    # belong to XSCompositePrefetcher in Prefetcher.py.
+    prefetcher.short_stride_thres = options.short_stride_thres
+
```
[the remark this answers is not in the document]
XXtaoo: This is defined by argparse as --short-stride-thres in Options.py; argparse maps it to options.short_stride_thres, so the direct access is valid.

Ergou-ren (approved):

PR #914, opened 2026-06-17
[codex] arch-riscv: Fix O3 fflags sticky updates

jensen-yan:
## Motivation

`povray_13677` in the SMT 0.3c perf path exposed a RISC-V difftest mismatch where `fflags` read back as `NX` in gem5 while the reference had `UF|NX`.

The root cause is that FP/vector FP instructions treated sticky `fflags` as a normal misc CSR: they read architectural `MISCREG_FFLAGS` at execute time and recorded a whole-value misc write for commit. On O3, younger FP instructions may execute before older ones, so a younger whole-value `fflags` write can overwrite sticky bits that an older FP instruction committed first.

## Approach

Port the upstream-style helper-register fix from gem5's `arch-riscv: Include FFLAGS O3 CPU fix from #868` direction:

- Add `MISCREG_FFLAGS_EXE` as an execution helper misc register outside the physical misc register file.
- Change scalar FP and vector FP templates to write only the current instruction's `softfloat_exceptionFlags` to `MISCREG_FFLAGS_EXE`.
- Handle `MISCREG_FFLAGS_EXE` in RISC-V ISA `setMiscReg()` by OR-ing the exception bits into architectural `MISCREG_FFLAGS` at commit time.
- Keep CSR writes to `fflags` / `fcsr` as whole-value writes, so explicit software clear/write semantics are unchanged.

## Validation

- `scons build/RISCV/gem5.fast --gold-linker -j64`

Warnings were limited to existing optional dependency warnings for PNG, HDF5, and backtrace support.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened floating-point exception flag validation and handling in RISC-V instruction execution.
  * Improved control/status register legality checks to more strictly validate out-of-range values.

* **Refactor**
  * Updated floating-point exception flag capture mechanism for improved accuracy across scalar and vector instructions.

### DIFF
--- a/src/arch/riscv/isa.cc
+++ b/src/arch/riscv/isa.cc
@@ -113,7 +113,7 @@ matrixWriteBlob(ThreadContext *tc, Addr addr, const void *src, size_t size)
 
 } // namespace
 
-[[maybe_unused]] const std::array<const char *, NUM_MISCREGS> MiscRegNames = {{
+[[maybe_unused]] const std::array<const char *, NUM_MISC_AND_HELPER_REGS> MiscRegNames = {{
     [MISCREG_PRV]           = "PRV",
     [MISCREG_VIRMODE]         = "VIRTUALIZATIONMODE",
     [MISCREG_ISA]           = "ISA",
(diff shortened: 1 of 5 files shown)


jensen-yan:
https://github.com/gem5/gem5/pull/868  此修复参考上游PR

PR #929, opened 2026-06-29
arch-riscv: calibrate RVV split rulses

jueshiwenli:
1. If a vector instruction other than `VectorUnitStrideLoadOp` is currently being split, all subsequent vector instructions must be stalled and cannot be issued.
2. If the vector instruction currently being split is a `VectorUnitStrideLoadOp`, other vector instructions may continue to enter the vector split unit without being blocked.
3. The issue behavior of non-vector instructions must not be inadvertently affected by this rule.

Change-Id: I3972e6b8191e2bb45007060d6492b741d93ec205

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved scheduling for vector memory operations so they become ready and issue more reliably.
  * Fixed cases where vector instructions could be delayed, duplicated, or left waiting incorrectly after completion, squashes, or commits.
  * Enhanced handling of vector split work so outstanding operations no longer interfere with new ones.
  * The issue queue now better reflects active vector split activity, improving overall readiness tracking.

### DIFF
--- a/src/cpu/o3/issue_queue.cc
+++ b/src/cpu/o3/issue_queue.cc
@@ -386,6 +386,29 @@ IssueQue::isVectorMemInst(const DynInstPtr& inst) const
     return inst && inst->isVector() && inst->isMemRef() && !inst->isSquashed();
 }
 
+bool
+IssueQue::isBlockingVectorSplitInst(const DynInstPtr& inst) const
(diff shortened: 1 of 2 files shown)


weidingliu (approved):
LGTM  :D

PR #934, opened 2026-06-30
cpu-o3: Split redirect to fetch delay

jensen-yan:
## Motivation

`commitToFetchDelay` currently has two meanings in the O3 frontend path:

- retire/commit feedback that Fetch consumes, such as decoupled BPU/FTQ commit release via `doneFtqId`, empty ROB notification, and interrupt feedback
- backend redirect/squash feedback that Fetch consumes before calling the BPU squash/restart path

That makes it easy to tune flush penalty by increasing `commitToFetchDelay`, but the model then also delays non-redirect commit feedback. RTL keeps the backend redirect path separate from FTQ commit release, so this patch splits the simulator parameter surface in the same direction.

## Approach

- Add `redirectToFetchDelay`, defaulting to `commitToFetchDelay` for compatibility with existing configs.
- Add a separate Fetch time-buffer wire for backend redirect feedback.
- Keep `doneFtqId` commit release on `commitToFetchDelay`.
- Consume squash/redirect payloads, `squashVersion`, and `robSquashing` through `redirectToFetchDelay`.
- Set KMHV3 and ideal-KMHV3 to `commitToFetchDelay = 2` and `redirectToFetchDelay = 4`.
- Restore ideal-KMHV3 `fetchToDecodeDelay` from 5 to 3 so fetch-to-decode latency is no longer used to carry extra flush penalty.

## Validation

- `python3 -m py_compile src/cpu/o3/BaseO3CPU.py configs/example/kmhv3.py configs/example/idealkmhv3.py`
- `git diff --check`
- `scons build/RISCV/gem5.opt --gold-linker -j64`
- `flush_penalty_small` with ideal-KMHV3 and commit-update predictors:
  - default config: `commitToFetchDelay=2`, `fetchToDecodeDelay=3`, `redirectToFetchDelay=4`
  - output: `measured_cycles=3406`, `branch_count=200`, about 17.03 cycles/miss
  - override `-P system.cpu[0].redirectToFetchDelay=2`
  - output: `measured_cycles=3006`, `branch_count=200`, exactly 2 cycles/miss lower

Follow-up cross-check:

- KMHV3 was initially much less sensitive to `redirectToFetchDelay` (`3010` vs `3006` cycles for 4 vs 2).
- That difference is explained by predictor update timing: current KMHV3 explicitly enables `resolvedUpdate` for MBTB/TAGE/ITTAGE, while this branch's ideal-KMHV3 still inherits commit-update defaults.
- Enabling `resolvedUpdate` in ideal-KMHV3 makes it behave like KMHV3 (`3010` vs `3006` cycles for 4 vs 2).
- Disabling `resolvedUpdate` in KMHV3 makes it behave like the commit-update ideal run (`3406` vs `3006` cycles for 4 vs 2).

So the new delay split behaves as expected; the microbench's apparent sensitivity also depends on whether the BPU learns from the resolve path before the measured B-chain redirect would otherwise be consumed.

### DIFF
--- a/configs/example/idealkmhv3.py
+++ b/configs/example/idealkmhv3.py
@@ -31,10 +31,11 @@ def setKmhV3IdealParams(args, system):
         cpu.fetchWidth = 32
         cpu.iewToFetchDelay = 2 # for resolved update, should train branch after squash
         cpu.commitToFetchDelay = 2
+        cpu.redirectToFetchDelay = 4
         cpu.fetchQueueSize = 64
 
         # decode
-        cpu.fetchToDecodeDelay = 5
+        cpu.fetchToDecodeDelay = 3
         cpu.decodeWidth = 8
         cpu.enable_loadFusion = False
         cpu.enableConstantFolding = False
--- a/configs/example/kmhv3.py
+++ b/configs/example/kmhv3.py
@@ -30,6 +30,7 @@ def setKmhV3Params(args, system):
         cpu.fetchWidth = 32
         cpu.iewToFetchDelay = 2 # for resolved update, should train branch after squash
         cpu.commitToFetchDelay = 2
+        cpu.redirectToFetchDelay = 4
         cpu.fetchQueueSize = 64
 
         # decode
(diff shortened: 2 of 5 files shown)


Yakkhini (approved):

jensen-yan:
idealkmhv3 性能正常，但kmhv3 版本性能异常，从18分降低到11分，暂时revert 这个PR 等待修复。

PR #991, opened 2026-07-27
Add Virtual SQ

happy-lx:
## Summary

This PR adds a VirtualSQ model to the gem5 O3 CPU. The model separates the
logical store-queue capacity used for dispatch from the physical SQ storage
capacity used when a store actually materializes its address or data.

For the KMHV3 configuration, the physical SQ is configured with 64 entries and
`StoreQueueMultiple=2`, giving a logical VirtualSQ capacity of 128 entries.
This allows dispatch to admit more stores than the physical SQ can currently
hold, while retaining the physical-SQ backpressure that affects store issue,
replay, and progress.

## Modeling Approach

The gem5 model intentionally does **not** mean to reproduce the RTL implementation
. It models the externally visible performance behavior with a simpler abstraction:

* `SQEntries` is the physical SQ capacity.
* The logical SQ capacity is `SQEntries * StoreQueueMultiple` and is used by
  dispatch/queue admission.
* A store STA or STD uop may be dispatched while the logical SQ has space even
  when the physical SQ window is full.
* When the STA or STD uop is issued and attempts to write the SQ, gem5 checks
  the physical capacity (`SQEntries`). If no physical entry is available, the
  uop is replayed and waits until the physical window can accept it.
* Physical-SQ replay reuses the existing memory replay path (`getPhysicalSQFullReplayInstToExecute()`).
  `phySQFullCheckAtReplay` controls whether a replay is reinserted immediately
  or waits in the `physicalSQFullReplayQs` until physical space is available.
* STA and STD share the same physical-window rule.

The physical-window check uses the monotonic SQ indices and an O(1) distance
check, so it does not scan or duplicate the RTL's physical-entry mapping logic.
The model also tracks physical-SQ replay and replay-blocked events separately
to make the source of backpressure visible in stats.

## Relation to RTL

In RTL, VirtualSQ separates logical pointer allocation from physical SQ storage.
Multiple logical positions can map to the same physical position, but a store
cannot update that physical position until the older occupant has been
dequeued. The resulting behavior is a replay/backpressure condition, even
though the store has already passed dispatch.

gem5 captures this same performance contract: dispatch can exceed the physical
SQ occupancy, but store execution cannot. The implementation is deliberately
behavioral rather than a cycle-accurate copy of RTL pointer updates, physical
entry reuse, or combinational arbitration. This keeps the hot path simple while
preserving the important causal chain:

`larger logical window -> more stores dispatched -> physical-SQ contention ->
store replay/waiting -> changed queue occupancy and IPC`

The design rationale and the corresponding RTL mechanism are described in
[architecture-committee PR #3](https://github.com/OpenXiangShan/architecture-committee/pull/3).

## Configuration and Observability

* `StoreQueueMultiple` must be a non-zero power of two; `1` preserves the
  previous behavior.
* `StoreQueueMultiple > 1` currently requires independent SMT LSQ mode.
* `storeReplayPhysicalSQFull` counts stores that replay because the physical
  SQ window was unavailable.
* `storePhysicalSQReplayBlocked` counts replay attempts that remain blocked by
  the physical window under the wait-before-replay policy.

## Summary by CodeRabbit

* **New Features**
  * Added configurable physical store-queue capacity and virtual-to-physical queue scaling.
  * Added controls for handling store-queue capacity limits during instruction replay.
  * Updated example KMHV3 configurations with explicit store-queue scaling.

* **Bug Fixes**
  * Improved handling of stores and atomic operations when physical store-queue space is unavailable.
  * Prevented replayed instructions from overtaking older in-flight instructions.
  * Enhanced queue draining and replay scheduling for more consistent execution.


jueshiwenli (approved):
Great! But I remember that Weiding’s configuration used a different SQ capacity from the current default, right?

happy-lx:
> Great! But I remember that Weiding’s configuration used a different SQ capacity from the current default, right?

image

Weiding master use 64 * 2 in RTL, I'm not sure if I've understood this correctly @weidingliu.

weidingliu:
This PR is awesome！ The `Virtual StoreQueue` is  64X2 in RTL, because the timing and performance of 48X2 is worse than 64X1. I will optimize 64X2 in the future.

weidingliu (approved):
LGTM ：）

happy-lx:
> This PR is awesome！ The `Virtual StoreQueue` is 64X2 in RTL, because the timing and performance of 48X2 is worse than 64X1. I will optimize 64X2 in the future.

 `Virtual StoreQueue` 非常好提案，使我醍醐灌顶
webwxgetmsgimg

PR #1027, opened 2026-08-07
cpu-o3: Add optional Preg SMT DynamicBorrowing policy

YihengChen1:
## Motivation

Under SMT, the default Dynamic Preg policy allows one thread to monopolize all physical registers (IntReg=224, FloatReg=192, VecReg=192) in the unified free list, starving the sibling thread. When a thread encounters heavy branch mispredictions or long-latency operations, its in-flight instructions hold physical registers indefinitely, blocking the other thread from renaming new instructions — causing extreme IPC imbalance (up to 172:1 in gcc_expr2 workloads).

Meanwhile, the Partitioned policy guarantees fairness but sacrifices ~2.7% overall throughput by preventing one thread from utilizing idle registers when the other thread has no demand.

## Approach

Introduce a **DynamicBorrowing** policy for SMT physical register allocation that dynamically adjusts per-thread register limits based on sibling thread activity. The core idea: when the sibling is idle (donor), allow borrowing most of its registers; when the sibling is active, protect its fair share.

**Donor determination** (per-cycle, before any thread renames):

```
isDonor(tid) = !hasPregDemand(tid) || backendBackpressureDonor(tid)
```

- `hasPregDemand`: checks fixedbuffer for non-squashed instructions needing dest registers
- `backendBackpressureDonor`: thread stalled on ROB-full (not Preg-full), with hold-cycle hysteresis to avoid frequent toggling

**Admission check** — per register type (Int/Float/Vec) independently:

```
borrowingLimit(type, tid):
    reserved = 0
    for other in all_threads:
        if other == tid: continue
        reserve = donor[other] ? donorQuota(type) : base(type)
        reserved += max(threadUsed[other][type], reserve)
    return max(numPhysRegs[type] - reserved, 0)

canAllocate(type, tid, n):
    return threadUsed[tid][type] + n <= borrowingLimit(type, tid)
```

Where:
- `base(type)` = `fixedBase > 0 ? min(fixedBase, numPhysRegs[type]/2) : numPhysRegs[type]/2`
- `donorQuota(type)` = `min(donorReserveRegs, base(type))`

**Key properties**:
- Default `smtPregPolicy` is `Dynamic`, which preserves the original full-share behavior with zero performance difference — existing configurations are unaffected
- `donorReserveRegs=0` → degenerates to Dynamic (no limit)
- `donorReserveRegs=numPhysRegs/2` → degenerates to Partitioned
- Donor state is determined uniformly for ALL threads before any renaming begins, preventing order-dependent behavior
- The `fixedBase` parameter enables a Watermark variant (both donor and non-donor use fixed thresholds)

## Environment

- **Branch**: `xs-dev`
- **Base commit**: `6d3120361e` (mem: fix bug of vsq (#1019))
- **Config**: SMT-2, KMHV3 scheduler, StoreQueueMultiple=1
- **Test**: SPEC06 INT, 12 workloads, 684 SimPoint checkpoints

## Results

| Config | score/GHz | vs Dynamic |
|--------|-----------|------------|
| Dynamic (baseline) | 23.799 | — |
| **DynamicBorrowing r=48** | **23.940** | **+0.59%** |
| DynamicBorrowing r=64 | 23.905 | +0.44% |
| Watermark d64_b64 | 23.834 | +0.15% |
| Partitioned | 23.161 | -2.68% |

Per-workload highlights (DynBorrow r=48):
- **mcf**: +1.95%
- **gcc**: +1.30%
- **sjeng**: +0.82%
- **omnetpp**: +0.68%

All 12 INT workloads show positive or neutral improvement with r=48. The sweet spot for donorReserveRegs is 48~64.

## Changes

8 files changed, +447 -47:
- `src/cpu/o3/BaseO3CPU.py` — add `smtPregPolicy` (default: `Dynamic`, preserving original full-share behavior), `smtPregFixedBase`, `smtPregBorrowDonorReserveRegs`, `smtPregBackendBackpressureDonor`, `smtPregBackendBackpressureDonorHoldCycles` parameters
- `src/cpu/o3/free_list.hh` — add DynamicBorrowing members (`pregPolicy`, `donor[]`, `threadUsed[][]`, `numPhysRegs[]`), public API (`canAllocate`, `borrowingLimit`, `setBorrowingDonor`, `isPerThreadExhausted`)
- `src/cpu/o3/free_list.cc` — implement `borrowingLimit()`, `canAllocate()`, `base()`, `donorQuota()`, allocation/free tracking per thread per type
- `src/cpu/o3/rename.hh` — add donor determination members (`pregBackendDonorCycles[]`, `hasPregDemand()`, `hasBackendQuotaFullStall()`)
- `src/cpu/o3/rename.cc` — per-cycle donor determination logic, DynamicBorrowing stall integration, stall reason reporting
- `src/cpu/o3/rename_map.hh` — extend `canRename()` to accept `ThreadID` for per-thread quota check
- `src/cpu/o3/rename_map.cc` — forward `canRename(tid)` to `freeList->canAllocate(type, tid)`
- `src/cpu/o3/cpu.cc` — pass Preg policy parameters from SimObject to FreeList/Rename

## Summary by CodeRabbit

* **New Features**
  * Added configurable physical-register sharing for simultaneous multithreading (SMT).
  * Supports fixed per-thread quotas, fair-share donor reserves, and policy-based allocation.
  * Threads can temporarily borrow available registers when backend pressure limits another thread.
  * Added controls for donor classification and configurable donor-status hold duration during backpressure.
  * Register allocation now adapts to active threads and tracks per-thread capacity.

* **Monitoring**
  * Added tracking for per-thread register exhaustion, register-full stalls, and donor activity.

### DIFF
--- a/src/cpu/o3/BaseO3CPU.py
+++ b/src/cpu/o3/BaseO3CPU.py
@@ -272,6 +272,20 @@ class BaseO3CPU(BaseCPU):
         8, "Cycles to keep an SMT thread marked as a ROB borrowing donor")
     smtBorrowDonorReserveEntries = Param.Unsigned(
         8, "Minimum ROB entries reserved for a borrowing donor to resume")
+    smtPregPolicy = Param.SMTQueuePolicy('Dynamic',
+                                         "SMT Preg (physical register) Sharing Policy")
+    smtPregFixedBase = Param.Unsigned(0,
+        "Fixed per-thread base quota for DynamicBorrowing (0 = numPhysRegs/activeThreads)")
+    smtPregBorrowDonorReserveRegs = Param.Unsigned(
+        8, "Minimum physical registers reserved for a borrowing donor to resume")
+    smtPregBackendBackpressureDonor = Param.Bool(True,
+        "Also treat a thread as a Preg borrowing donor when it is stalled "
+        "on ROB/dispatch-queue-bandwidth backpressure (a resource other "
+        "than Preg itself); when false, only actual Preg demand this "
+        "cycle drives donor status")
+    smtPregBackendBackpressureDonorHoldCycles = Param.Unsigned(
+        8, "Cycles to keep a backend-backpressure-triggered Preg donor "
+           "marking held after the triggering condition clears")
     smtBorrowBaseReserveEntries = Param.Unsigned(
         80, "Minimum ROB entries reserved for a borrowing base to resume")
 
--- a/src/cpu/o3/BaseO3CPU.py
+++ b/src/cpu/o3/BaseO3CPU.py
@@ -276,8 +276,9 @@ class BaseO3CPU(BaseCPU):
                                          "SMT Preg (physical register) Sharing Policy")
     smtPregFixedBase = Param.Unsigned(0,
         "Fixed per-thread base quota for DynamicBorrowing (0 = numPhysRegs/activeThreads)")
-    smtPregBorrowDonorReserveRegs = Param.Unsigned(
-        8, "Minimum physical registers reserved for a borrowing donor to resume")
+    smtPregDonorReservePercent = Param.Unsigned(
+        43, "Percentage (0-100) of per-thread fair share reserved for a "
+            "donor thread.  donorQuota = numPhysRegs/activeThreads * pct/100")
     smtPregBackendBackpressureDonor = Param.Bool(True,
         "Also treat a thread as a Preg borrowing donor when it is stalled "
         "on ROB/dispatch-queue-bandwidth backpressure (a resource other "
--- a/src/cpu/o3/cpu.cc
+++ b/src/cpu/o3/cpu.cc
@@ -105,7 +105,7 @@ CPU::CPU(const BaseO3CPUParams &params)
               params.numPhysRMiscRegs,
               params.isa[0]->regClasses()),
 
-      freeList(name() + ".freelist", &regFile),
+      freeList(name() + ".freelist", &regFile, params),
 
       rob(this, params),
 
@@ -165,6 +165,7 @@ CPU::CPU(const BaseO3CPUParams &params)
     rename.setActiveThreads(&activeThreads);
     iew.setActiveThreads(&activeThreads);
     commit.setActiveThreads(&activeThreads);
+    freeList.setActiveThreads(&activeThreads);
 
     // Give each of the stages the time buffer they will use.
     fetch.setTimeBuffer(&timeBuffer);
@@ -242,8 +243,8 @@ CPU::CPU(const BaseO3CPUParams &params)
     for (ThreadID tid = 0; tid < numThreads; tid++) {
         isa[tid] = dynamic_cast<TheISA::ISA *>(params.isa[tid]);
         warn("Setting isa ptr of cpu to %p", isa[tid]);
-        commitRenameMap[tid].init(regClasses, &regFile, &freeList);
-        renameMap[tid].init(regClasses, &regFile, &freeList);
+        commitRenameMap[tid].init(regClasses, &regFile, &freeList, tid);
+        renameMap[tid].init(regClasses, &regFile, &freeList, tid);
     }
 
     // Initialize rename map to assign physical registers to the
@@ -256,7 +257,7 @@ CPU::CPU(const BaseO3CPUParams &params)
                 // Note that we can't use the rename() method because we don't
                 // want special treatment for the zero register at this point
                 RegId rid = RegId(type, ridx);
-                PhysRegIdPtr phys_reg = freeList.getReg(type);
+                PhysRegIdPtr phys_reg = freeList.getReg(type, tid);
                 renameMap[tid].setEntry(rid, VirtRegId(phys_reg));
                 commitRenameMap[tid].setEntry(rid, VirtRegId(phys_reg));
             }
@@ -643,6 +644,7 @@ CPU::startup()
     iew.startupStage();
     rename.startupStage();
     commit.startupStage();
+    freeList.resetEntries();
 }
 
 void
@@ -807,7 +809,7 @@ CPU::insertThread(ThreadID tid)
     for (auto type = (RegClassType)0; type <= RMiscRegClass;
             type = (RegClassType)(type + 1)) {
         for (RegIndex idx = 0; idx < regClasses.at(type).numRegs(); idx++) {
-            PhysRegIdPtr phys_reg = freeList.getReg(type);
+            PhysRegIdPtr phys_reg = freeList.getReg(type, tid);
             renameMap[tid].setEntry(RegId(type, idx), VirtRegId(phys_reg));
             scoreboard.setReg(phys_reg);
         }
@@ -825,6 +827,7 @@ CPU::insertThread(ThreadID tid)
 
     //Reset ROB/IQ/LSQ Entries
     commit.rob->resetEntries();
+    freeList.resetEntries();
 }
 
 void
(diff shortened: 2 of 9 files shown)


XDu5 (approved):

XDu5 (dismissed):

jensen-yan (changes requested):
Thanks for adding a Preg sharing policy between fully Dynamic and strictly Partitioned. The overall direction makes sense, and the reported throughput result is promising. I rechecked the current head (`02c6e0c5`); the following points still need to be addressed before merge:

1. **The `threadUsed` lifecycle is incomplete.** Every allocation, including the initial RAT mappings in the constructor and `insertThread()`, increments `threadUsed[tid][type]`. Normal commit/squash release decrements it, but `CPU::removeThread()` neither returns all mappings owned by the thread nor rebuilds/resets this ownership state. Reusing the same tid can therefore accumulate stale occupancy and eventually exhaust its Partitioned/DynamicBorrowing quota. Simply zeroing the counter is not sufficient unless the corresponding physical registers are also returned. Please make allocation/teardown/reinsertion accounting consistent, or explicitly reject the unsupported lifecycle. Assertions before the unsigned decrement would also help catch mismatched release paths.

2. **`base()` and `borrowingLimit()` use different thread sets.** `base()` divides by `activeThreadCount()`, while `borrowingLimit()` reserves capacity for every configured thread. The earlier SMT2 “single active thread always gets zero” report is not quite accurate once the inactive sibling is marked as a donor, but the mismatch is still unsafe for N-way SMT. For example, with four configured threads, one active thread, 192 Vec Pregs, and `donorReserveRegs=64`, three inactive donors reserve all 192 entries and the active thread gets a limit of zero. Please either use a consistent thread set, constrain this policy to SMT2, or validate the N-way parameter range so a legal configuration cannot deadlock rename.

3. **Backend-donor behavior does not match the exposed parameters.** `hasBackendQuotaFullStall()` recognizes only `ROBFull`, although the parameter/comments promise ROB and dispatch-queue bandwidth backpressure. Also, `HoldCycles=0` currently disables the trigger even in the triggering cycle. Please either implement the documented semantics (including Mem/Int/FV DQ bandwidth and current-cycle donation) or narrow the parameter descriptions.

Since the motivation is starvation/fairness (up to a 172:1 IPC ratio), please also report whether that worst case is actually improved, ideally with a per-thread IPC/fairness metric in addition to aggregate score/GHz. Targeted tests for SMT2 single-active, SMT4/N-way limits, donor transitions, and remove/reinsert would be valuable.

Non-blocking: the quota formula and donor bookkeeping are very similar to the existing ROB DynamicBorrowing logic. A shared policy/helper, or at least common invariants/tests, would reduce the risk that ROB and Preg semantics drift over time.

jensen-yan (dismissed):

YihengChen1:
### Issue 1: threadUsed Lifecycle

**Fixed.** Added `resetThreadUsed(ThreadID tid)` called from `CPU::removeThread()` after thread mappings are released, clearing per-thread counters and donor state. Also added an assert before decrement in `addReg()` to catch underflow. `borrowingLimit()` now iterates only the `activeThreads` list so inactive threads don't participate in reserve calculations.

---

### Issue 2: N-way SMT Scalability

**Fixed.** Replaced the absolute parameter `smtPregBorrowDonorReserveRegs` with percentage-based `smtPregDonorReservePercent` (0-100):

```cpp
unsigned UnifiedFreeList::donorQuota(RegClassType type) const {
    unsigned fairShare = numPhysRegs[type] / activeThreadCount();
    return fairShare * donorReservePercent / 100;
}
```

| Scenario | donorQuota |
|----------|-----------|
| SMT-2, IntReg=224, pct=43 | 112 × 43/100 = **48** |
| SMT-4, IntReg=224, pct=43 | 56 × 43/100 = **24** |

The percentage parameter scales automatically with thread count, preventing the sum of reserves from exceeding the total register file.

---

### Issue 3: hasBackendQuotaFullStall Coverage

**Implemented per documented semantics**, covering ROBFull + Mem/Int/FV DQ Bandwidth, with the HoldCycles=0 current-cycle donation fix:

```cpp
bool Rename::hasBackendQuotaFullStall(ThreadID tid) {
    switch (checkRenameStallFromIEW(tid)) {
      case StallReason::ROBFull:
      case StallReason::MemDQBandwidth:
      case StallReason::IntDQBandwidth:
      case StallReason::FVDQBandwidth:
        return true;
      default:
        return false;
    }
}
```

HoldCycles=0 fix — donation takes effect on the triggering cycle:
```cpp
bool backendBackpressureDonor = backendQuotaFull || pregBackendDonorHoldRemaining[i] > 0;
```

Also renamed member `pregBackendDonorCycles` to `pregBackendDonorHoldRemaining` to avoid confusion with the identically-named stats counter.

**Experimental data:** `pregBackendDonorCycles = 0` across all 684 slices — the backend-backpressure donor never fires.

**Reason:** Under the current resource ratio (IntReg=224, SMT-2), Preg is a tighter resource than ROB. The `hasPregDemand` criterion always triggers before backend backpressure — by the time a thread stalls due to Preg exhaustion, its ROB is far from full. **Under the current configuration, the only effective donor criterion is `!hasPregDemand(tid)`.**

`hasBackendQuotaFullStall` serves as a reserved fallback mechanism for when Preg resources are expanded in the future such that `hasPregDemand` alone cannot effectively identify donors. This policy is a validated positive-gain configuration under current resource constraints, but does not represent the optimal configuration.

---

### Issue 4: Inter-thread IPC Imbalance

**Correction of prior PR wording:** The earlier PR description incorrectly expressed the Preg full events ratio as an IPC ratio, causing confusion. `fullRegistersEvents` (Preg Full Events) is the correct metric for measuring per-thread Preg fairness.

**Preg Full Events data** (extreme slice `gcc_expr2_6943`):

| Mode | t0 | t1 | Ratio |
|------|----|----|-------|
| Dynamic | 4,942,274 | 9,256 | 99.8% : 0.2% |
| DynamicBorrowing | 373 | 46 | 89% : 11% |

Under Dynamic mode, Preg full events are extremely imbalanced (t0 suffers 99.8%). DynamicBorrowing reduces the total from 4.95M to 419 (↓99.99%) and brings the inter-thread ratio closer to balanced, confirming the policy effectively mitigates Preg-level unfairness.

**IPC ratio change:**

| Metric | Dynamic | DynamicBorrowing |
|--------|---------|-----------------|
| t0 IPC | 0.781 | 1.142 (+46%) |
| t1 IPC | 0.581 | 0.712 (+22%) |
| IPC ratio (t0/t1) | 1.34 | 1.60 |
| **Total throughput (IPC sum)** | **1.362** | **1.854 (+36%)** |

While Preg full events are now balanced, the IPC ratio still widens from 1.34 to 1.60. This indicates that **the IPC ratio expansion is not caused by Preg allocation unfairness**, but likely results from imbalanced allocation of other shared resources (e.g., IQ contention, ROB occupancy). Once t0's Preg bottleneck is removed, t0 runs faster and consumes more of other shared resources, indirectly limiting t1's improvement. This is a systemic SMT shared-resource issue, not one introduced by the Preg policy alone.

---

### Issue 5: Reusing ROB Donor State

**Experiment:** Created a variant where Preg donor directly reads the ROB's existing `borrowingDonor` state:

```cpp
bool isDonor = commit_ptr->rob->isBorrowingDonor(i);
freeList->setBorrowingDonor(i, isDonor);
```

**Results:**

| Configuration | SPEC INT Score |
|---------------|---------------|
| Independent Preg donor (`!hasPregDemand`) | **23.940** |
| Reusing ROB donor | 23.826 (-0.48%) |

**Analysis:** Directly reusing the ROB donor policy results in a slight performance regression. The ROB donor trigger condition (backend throttle stall + holdCycles) is more conservative than `!hasPregDemand` — ROB identifies threads that are "severely backend-stalled", while `hasPregDemand` identifies threads that "currently have no register demand", which is more responsive and has broader coverage. Due to the performance regression, we have not adopted the identical strategy and retain the independent Preg donor logic for now.

Future work may explore more refined donor identification mechanisms.

jensen-yan (approved):

PR #1029, opened 2026-08-07
util: Update RVV CI set

jueshiwenli:
## Summary by CodeRabbit

* **New Features**
  * RVV performance testing can now be triggered for labeled pull requests from external forks targeting `xs-dev`.
  * Added an option to explicitly permit external pull-request code during performance workflow runs.

* **Documentation**
  * Clarified label support for same-repository and external-fork pull requests.
  * Added FAQ guidance about external pull-request handling and self-hosted runner execution.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -58,7 +58,7 @@
 
 - 添加 `perf` / `perf-align` 标签会在对应性能 workflow 中触发测试，workflow 会记录标签创建时 PR 的 head SHA 确保结果对应正确的 commit
 - `rvv` 标签仍由独立的 RVV on-demand workflow 触发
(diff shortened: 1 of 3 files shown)


jensen-yan:
Thanks for the update. After reviewing the external-fork path again, I think a label is a convenient maintainer approval mechanism for on-demand CI, but this implementation would deliberately bypass the fork protection in `actions/checkout@v7` and execute the external PR head on a trusted self-hosted runner.

The current patch would also need additional checkout handling to work reliably: `allow-unsafe-pr-checkout` must be explicitly enabled and the fork repository / immutable PR head SHA must be selected correctly.

For now, we will keep external-fork performance tests maintainer-triggered through `manual-perf.yml`. If this becomes a common workflow, we can revisit it with a dedicated PR-number/head-SHA input, stricter authorization or runner isolation, and automatic posting of the run link/result back to the PR.

Closing this for now; we can reopen or replace it once that trust model is defined.

PR #1032, opened 2026-08-11
ci: Add manual H SPEC06 performance workflow

Gekyume777:
Add 0.5c and 1.0c H SPEC06 workflow-dispatch variants. The default configuration is idealkmhv3 with DRAMsim3, the FS0 H NEMU reference, and the GCPT restorer required by H checkpoints.

Use static checkpoint lists and H profile instruction-count JSONs so checkpoint selection and score normalization are reproducible. Preserve partial checkpoint status and artifacts after a failed batch, while validation still marks incomplete runs as failed.

## Summary by CodeRabbit

* **New Features**
  * Added H SPEC06 performance benchmark options for 0.5c and 1.0c configurations.
  * Added tools to generate checkpoint lists, run distributed benchmarks, score results, and validate reports.
  * Added manual controls for benchmark selection, coverage, prefetch profiles, revisions, and distributed execution.
  * Added CSV, JSON, and text score reports with workload-level and overall performance results.
  * Added validation for required files, benchmark selections, execution results, and score summaries.


jensen-yan (changes requested):
Thanks for adding the H SPEC06 workflow. Before merging, please try to reuse the existing `gem5-perf-template.yml` as much as possible instead of introducing a parallel execution template.

The new H template duplicates most of the existing CI lifecycle: checkout and trust checks, DRAMSim/gem5 build, archive and metadata handling, distributed execution, scoring, summaries, and artifact upload. The H-specific differences (checkpoint root/list, NEMU reference, restorer, fixed gem5 arguments, and stricter result validation) should preferably be expressed as benchmark-specific configuration or delegated to the H helper script while keeping the common workflow in one template. Otherwise, future changes to checkout security, runner configuration, retry policy, archive layout, or scoring will need to be kept synchronized across two templates, which is difficult to maintain and violates DRY. A separate thin manual entry workflow is fine, but the underlying execution template should remain shared where possible.

Also, according to the current CI convention, checkpoint selection JSON and list files are normally stored under the shared server checkpoint/profile data directory rather than committed to the repository. Keeping them in the repository can work for now.

Please also address the existing actionable review comments before this is merged.

Gekyume777:
> Thanks for adding the H SPEC06 workflow. Before merging, please try to reuse the existing `gem5-perf-template.yml` as much as possible instead of introducing a parallel execution template.
> 
> The new H template duplicates most of the existing CI lifecycle: checkout and trust checks, DRAMSim/gem5 build, archive and metadata handling, distributed execution, scoring, summaries, and artifact upload. The H-specific differences (checkpoint root/list, NEMU reference, restorer, fixed gem5 arguments, and stricter result validation) should preferably be expressed as benchmark-specific configuration or delegated to the H helper script while keeping the common workflow in one template. Otherwise, future changes to checkout security, runner configuration, retry policy, archive layout, or scoring will need to be kept synchronized across two templates, which is difficult to maintain and violates DRY. A separate thin manual entry workflow is fine, but the underlying execution template should remain shared where possible.
> 
> Also, according to the current CI convention, checkpoint selection JSON and list files are normally stored under the shared server checkpoint/profile data directory rather than committed to the repository. Keeping them in the repository can work for now.
> 
> Please also address the existing actionable review comments before this is merged.

Thanks for the detailed review. Agreed. I will refactor the H workflow to reuse gem5-perf-template.yml for the shared CI lifecycle, and keep only H-specific checkpoint/ref/restorer/validation behavior in a thin entry workflow and helper layer. I will also address the actionable comments, run a smoke validation, and update this PR.

jueshiwenli (approved):

PR #1041, opened 2026-08-13
ci: remove distributed SSH dispatch host

jensen-yan:
## What changed

- route distributed performance and solver workloads directly from the self-hosted runner to worker nodes
- remove the `--dispatch-host` interface and two-hop SSH wrapper from the distributed runner, H runner, and solver executor
- remove the hardcoded `ci-runner@172.28.9.101` workflow configuration and stale metadata

## Why

The self-hosted runners and node020-node039 are in the same machine room. Batch-mode tests from the `cirunner` account successfully connected directly to node020, node021, node029, node030, node038, and node039. Routing every probe and launch through 172.28.9.101 adds an unnecessary cross-room hop and makes that host a single point of failure.

The failing SMT run 31688035148 showed the issue: the job ran on node029, but even probing node029 was sent through 101 and timed out.

## Impact

Distributed CI now uses the existing direct SSH path and preserves idle probing, launch retries, bad-node handling, SSH options, server-domain, ssh-config, and ssh-user support.

## Validation

- Python syntax compile for the four modified Python files
- YAML parse for both modified workflows
- CLI help checks for `distributed_sim.py` and `h_spec06_perf.py`
- repository search confirms no remaining dispatch-host/101 references in the affected call chain
- direct BatchMode SSH confirmed by the CI `cirunner` account on node020, node021, node029, node030, node038, and node039

## Summary by CodeRabbit

* **Changes**
  * Distributed workloads now connect directly to worker servers using the configured SSH settings.
  * Removed support for configuring a separate dispatch host from distributed simulation and solver commands.
  * Existing distributed SSH options and execution behavior remain available without dispatch-host routing.

### DIFF
--- a/.github/workflows/gem5-perf-template.yml
+++ b/.github/workflows/gem5-perf-template.yml
@@ -265,7 +265,6 @@ jobs:
           CONFIG_NAME=${CONFIG_NAME:-config}
           TARGET_DIR="${BENCHMARK_DIR}/${TIMESTAMP}_${COMMIT_SHORT}_${CONFIG_NAME}_run${RUN_NUMBER}"
           DEFAULT_DISTRIBUTED_SERVERS="node020-node034,node036-node039"
-          DISTRIBUTED_DISPATCH_HOST="ci-runner@172.28.9.101"
           DISTRIBUTED_SERVERS_INPUT="$DISTRIBUTED_SERVERS_RAW"
           DISTRIBUTED_SERVERS=$(echo "$DISTRIBUTED_SERVERS_INPUT" | xargs)
           if [ "$DISTRIBUTED_SERVERS" = "default" ]; then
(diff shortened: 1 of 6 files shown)


jensen-yan:
## Distributed CI validation

The dedicated validation run is in progress: [SMT SPEC2006 0.3c run #137](https://github.com/OpenXiangShan/GEM5/actions/runs/31693081884).

Observed so far:

- the job is running on self-hosted runner `node029-1` and is using the workflow/template from the test branch containing this PR
- direct idle probes succeeded for `node029`, `node030`, `node032`, `node033`, `node034`, `node037`, and `node039`
- `node036` and `node038` were skipped only because they did not meet the required 32 idle physical cores, rather than because of an SSH failure
- the scheduler selected all 148 workloads and began distributing them across the seven usable nodes
- the observed log contains no `172.28.9.101`, dispatch-host hop, SSH timeout, or `exit=255`

The full performance run is still ongoing, so this is not yet an end-to-end performance completion result. However, the behavior relevant to this PR is working: the runner can directly probe the worker pool and initiate distributed jobs without relying on the 101 server.

PR Quick Check also passed. This looks safe to merge ahead of tonight's weekly run while the dedicated validation continues in the background.

PR #1047, opened 2026-08-14
cpu: Split commit-squash stalls by squash cause

aphlta:
## Motivation

`dispatchStallReason::CommitSquash` lumps together every stall caused by a squash, so top-down analysis cannot tell *why* the pipeline is recovering. On coremark this single bucket accounts for ~10% of dispatch stall ticks (and up to ~27% on some workloads), making it one of the largest un-attributed items under Bad Speculation.

Commit already counts the individual causes (`squashDueToBranch`, `squashDueToOrderViolation`, ...), but that information never reaches the stage that records stall reasons.

## Implementation

IEW cannot derive the cause on its own: `squashAll()` clears `mispredictInst`, and no cause-bearing state is live during the multi-cycle `robSquashing` tail. Therefore:

- Add `enum class SquashCause` plus a per-thread latch in commit, set at every squash initiation point (branch mispredict, value-prediction error, memory-order violation, trap, thread-context write, squash-after). `TrapPending` sets it one cycle before `squashFromTrap()` runs, which is where the stall actually starts.
- Republish it through `CommitComm` on every squashing cycle, so IEW needs no latch of its own. `SquashCause::None` is kept at zero because `TimeBuffer::advance()` memsets new slots.
- Map it onto four new stall reasons: `ControlRecovery`, `MemVioRecovery`, `VPRecovery`, `TrapRecovery`. `CommitSquash` is retained as the fallback for simulator-side clears (TC writes, squash-after), so no existing stat name disappears and old/new runs remain comparable.
- Also drop an unreachable `return` at the end of `IEW::checkDispatchStall()`.

## Test plan

- `scons build/RISCV/gem5.opt` builds clean.
- `scons build/RISCV/unittests.opt --unit-test`: 90 tests from 6 suites passed.
- Coremark (2 iterations) with difftest against NEMU: passes, CoreMark CRCs correct, and IPC is bit-identical to the base branch (0.785744), confirming this is a pure attribution change.
- The new buckets agree with the pre-existing event counters:

| commit event counter | resulting stall bucket (dispatch) |
|---|---|
| `squashDueToBranch` = 2809 | `ControlRecovery` = 672929 ticks (9.96%) |
| `squashDueToOrderViolation` = 2 | `MemVioRecovery` = 160 ticks |
| `squashDueToTC` = 1 | `CommitSquash` = 1544 ticks (fallback, by design) |

## Summary by CodeRabbit

* **Performance and Diagnostics**
  * Improved processor pipeline stall reporting with distinct reasons for control-flow, memory-order, value-prediction, and trap recovery.
  * Recovery-related stalls are now classified more accurately in performance statistics.
  * Enhanced multi-threaded resource handling by preventing recovery stalls from being treated as shareable capacity.

### DIFF
--- a/src/cpu/o3/comm.hh
+++ b/src/cpu/o3/comm.hh
@@ -100,12 +100,53 @@ enum StallReason {
     VectorReadyButNotIssued,  // B
     ScalarReadyButNotIssued,  // B
     ResumeUnblock,  // B
-    CommitSquash,  // BS
+    CommitSquash,  // BS, cause not attributable
+    ControlRecovery,  // BS
+    MemVioRecovery,  // BS
+    VPRecovery,  // BS
+    TrapRecovery,  // BS
     ROBFull,  // B
     RegFull,  // B
     OtherStall,  // B
     NumStallReasons
 };
+
+/**
+ * Why commit squashed. IEW cannot derive this itself: `squashAll()` clears
+ * `mispredictInst`, and nothing cause-bearing is live during the
+ * `robSquashing` tail.
+ *
+ * `None` must stay zero; TimeBuffer::advance() memsets new slots.
+ */
+enum class SquashCause
+{
+    None = 0,
+    BranchMispredict,
+    MemOrderViolation,
+    ValuePrediction,
+    Trap,
+    ThreadContext,
+    SquashAfter,
+};
+
+inline StallReason
+squashCauseToStallReason(SquashCause cause)
+{
+    switch (cause) {
+      case SquashCause::BranchMispredict:
+        return StallReason::ControlRecovery;
+      case SquashCause::MemOrderViolation:
+        return StallReason::MemVioRecovery;
+      case SquashCause::ValuePrediction:
+        return StallReason::VPRecovery;
+      case SquashCause::Trap:
+        return StallReason::TrapRecovery;
+      // TC writes / squash-after are simulator-side clears, not
+      // microarchitectural events worth their own bucket.
+      default:
+        return StallReason::CommitSquash;
+    }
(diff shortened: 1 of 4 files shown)


jueshiwenli:
Hello, thank you for your contribution. Would it be possible for us to exchange contact information? My email address is [zhenhao@ict.ac.cn](mailto:zhenhao@ict.ac.cn).

aphlta:
Hello, I am happy to provide my mail. it's alexcoder2022@gmail.com.

aphlta:
Thanks for the review notes — follow-up pushed in 930e995.

**Interrupt `TrapPending` (fixed).**  
`curSquashCause` is now set to `SquashCause::Trap` inside `generateTrapEvent()`, which is the common path for both instruction-fault and interrupt-driven traps. The previous latch next to `commitHead()`'s `TrapPending` was removed so there is a single source of truth.

**SMT + `setAllStalls()` (not in this PR).**  
Agreed that `setAllStalls()` writes a shared `dispatchStalls` vector. That overwrite existed before this change (everything was labeled `CommitSquash`); splitting causes only makes a wrong label more visible under true SMT. Fixing it properly needs per-thread dispatch stall state and is out of scope here — happy to track that as a follow-up if maintainers want it.

**Naming (`squash_cause_to_stall_reason`).**  
Left as-is to match the surrounding o3 camelCase helpers (`setAllStalls`, `checkSquash`, …). Can rename if reviewers prefer snake_case for new symbols.

jensen-yan (approved):

jueshiwenli (approved):
Great, thanks for your contribution!

PR #1077, opened 2026-08-26
mem-cache: Avoid shifting negative strides

jensen-yan:
## Motivation

Extended UBSan runs after #1070 found two reports in the XiangShan-specific XSStride prefetcher:

```text
src/mem/cache/prefetch/xs_stride.cc:183:80: runtime error: left shift of negative value -16
src/mem/cache/prefetch/xs_stride.cc:185:80: runtime error: left shift of negative value -16
```

The issue reproduces with both gobmk and sjeng. XSStride is not present in upstream gem5, so there is no upstream fix to backport.

## Change

Compute the 4x and 32x lookahead addresses in the unsigned `Addr` domain. Converting the signed stride to `Addr` makes wraparound explicit and defined, preserving the existing backward-prefetch address semantics without signed left-shift UB. The same computed addresses are shared with ArchDB tracing.

## Validation

- `CC=clang CXX=clang++ scons build/RISCV/gem5.opt --gold-linker -j64 --with-ubsan --without-tcmalloc`
- gobmk_nngs, 250,000 instructions: XSStride UBSan reports removed
- sjeng, 250,000 instructions: XSStride UBSan reports removed
- Before/after execution is unchanged:
  - gobmk: `simTicks=94047525`, `simInsts=250001`, `numCycles=282426`
  - sjeng: `simTicks=70412517`, `simInsts=250006`, `numCycles=211450`

On the current `xs-dev` base, the remaining startup reports are limited to the upstream issues addressed separately by #1076.

### DIFF
--- a/src/mem/cache/prefetch/xs_stride.cc
+++ b/src/mem/cache/prefetch/xs_stride.cc
@@ -180,20 +180,29 @@ XSStridePrefetcher::strideLookup(AssociativeSet<StrideEntry> &stride, const Pref
             unsigned start_depth = pfi.isCacheMiss() ? std::max(1, (entry->depth - 4)) : entry->depth;
             Addr pf_addr = 0;
             if (useXsDepth) {
-                sendPFWithFilter(pfi, blockAddress(lookupAddr + (entry->stride << 2)), addresses, 0,
+                // Keep negative-stride address arithmetic in the unsigned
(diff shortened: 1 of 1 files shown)


eastonman:
I have met this before using ubsan, seems to be a good fix.

happy-lx (approved):

PR #1089, opened 2026-08-27
cpu: Store folded-history snapshots as values

jensen-yan:
## Motivation

`perf` on a fixed `gem5.opt` CoreMark run showed repeated folded-history metadata copies in the host hot path. TAGE and MGSC metadata copied full folded-history objects, including their per-shift position tables, although later lookup and recovery only consume the folded value.

## Approach

- Store TAGE and MGSC folded-history snapshots as `uint64_t` values.
- Add `recoverValue()` to restore only the folded value while retaining the live table configuration.
- Preserve the existing full guest history snapshot and database-trace hashes.
- Update the BTBTAGE unit test for the value metadata type.

No predictor parameter, table organization, guest-visible timing, or stats definition changes. The changed prediction/recovery path is present in both opt and fast builds; it is not a debug-only/NDEBUG path.

## Validation

Baseline: `f63979dc6c` (#1067), node102 CPU 4, `gem5.opt`, `configs/example/kmhv3.py`, raw CoreMark (10 iterations), difftest enabled.

- Guest exit tick: `473882310` for baseline and candidate in every run.
- Normalized `stats.txt`: no differences after filtering host-time fields.
- Preliminary interleaved screening: two stable adjacent pairs improved task-clock by 1.69% and 1.59%.
- Final single-commit adjacent pair: 30.471 s -> 29.745 s (+2.44%); retired instructions 319,906,979,297 -> 319,044,893,004 (-0.27%).
- `tage.test.debug`: 28/28 passed.
- `mgsc.test.debug`: 14/14 passed.
- `folded_hist.test.debug`: 10/10 passed.

PGO-fast and RVV full regressions were not run locally; CI can cover those paths.

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced overhead during branch prediction by storing compact history snapshots instead of copying full history structures.
  * Improved prediction-state recovery using lightweight value restoration.
* **Bug Fixes**
  * Improved consistency when restoring prediction histories across TAGE and MGSC components.
* **Tests**
  * Updated branch prediction tests to reflect the revised history-state handling.

### DIFF
--- a/src/cpu/pred/btb/btb_mgsc.cc
+++ b/src/cpu/pred/btb/btb_mgsc.cc
@@ -35,6 +35,26 @@ namespace branch_prediction
 namespace btb_pred
 {
 
+namespace
+{
+
+// Snapshot a folded-history vector as raw folded values. Prediction and tag
+// lookups read the values directly and recovery restores them through
+// recoverValue(), so the metadata only needs the folded value and not the full
+// object with its per-shift position tables.
+template <typename Src>
+void
+snapshotFoldedValues(const Src &src, std::vector<uint64_t> &dst)
+{
+    dst.resize(src.size());
+    for (size_t i = 0; i < src.size(); ++i) {
+        dst[i] = src[i].get();
+    }
+}
+
+}  // namespace
+
+
 #ifdef UNIT_TEST
 namespace test
 {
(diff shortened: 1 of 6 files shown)


jensen-yan:
感觉改动有点复杂，算了吧，不是很着急

PR #1098, opened 2026-08-31
cpu: Fix 128-bit AMO (Zacas amocas.q) difftest golden-read path

civil233:
The difftest golden-read path for AMO instructions was designed for
  64-bit (8-byte) AMOs. Zacas extension's amocas.q is a 128-bit (16-byte)
  CAS that overflows every buffer/field in the chain.

  Changes:
  - issue_golden[8]->[16] (lsq_unit.cc, 2 places)
  - tmp_data[8]->[16] (lsq_unit.cc completeStore AMO path)
  - goldenData[8]->[16] (dyn_inst.hh)
  - amoOldGoldenValue uint64_t->uint8_t[16] (dyn_inst.hh + base.hh)
  - readGoldenMem: direct memcpy bypassing pmemRead switch
  - Clamp memcmp/memcpy to prevent read-past on gem5_val (uint64_t)
  - Remove redundant panic_if guards (buffers now sized for 16 bytes)

  Verified: perlbench_02/6326 (was AMO-panic) -> PASS
           Full 353 regression(our spec17int-multicore slices):  0 AMO-panic , only 17/353 trap fail

## Summary by CodeRabbit

* **New Features**
  * Added support for 128-bit atomic memory operations.
  * Extended atomic operation tracking and memory updates to handle 16-byte values.
  * Improved validation for memory ranges spanning multiple bytes.

* **Bug Fixes**
  * Improved atomic-operation verification and recovery against the reference system.
  * Prevented invalid memory reads by safely clearing data when a requested range falls outside physical memory.

### DIFF
--- a/src/cpu/base.cc
+++ b/src/cpu/base.cc
@@ -1530,6 +1530,102 @@ BaseCPU::diffWithNEMU(ThreadID tid, InstSeqNum seq)
             DPRINTF(Diff, "At %s Ref value: %#lx, GEM5 value: %#lx\n",
                     reg_name[dest_tag], nemu_val, gem5_val);
 
+            // 128-bit AMO (amocas.q): handle both dest regs as a pair
+            // at dest_idx=0, then skip dest_idx=1 to handle the 128bit result as one pair.
+            if (diffInfo.effSize > 8 && diffInfo.inst->isAtomic() &&
+                dest_idx == 0 && diffInfo.inst->numDestRegs() >= 2) {
+
+                // high 64bit of 128bit AMO load
+                const auto &dest1 = diffInfo.inst->destRegIdx(dest_idx + 1);
+                auto dest1_tag = dest1.index() + dest1.isFloatReg() * 32;
+                uint64_t gem5_hi = diffInfo.scalarResults[dest_idx + 1];
+                uint64_t nemu_hi = diffAllStates->referenceRegFile[dest1_tag];
+
+                DPRINTF(Diff, "128-bit AMO pair: lo GEM5=%#lx REF=%#lx, "
+                        "hi GEM5=%#lx REF=%#lx\n",
+                        gem5_val, nemu_val, gem5_hi, nemu_hi);
+                if (gem5_val == nemu_val && gem5_hi == nemu_hi) {
+                    // skip dest_idx=1
+                    dest_idx++;
+                    continue;
+                }
+
+                // if not completely equal, need to be repaired or set diff_at
+                bool repaired = false;
+                if (system->multiContextDifftest() &&
+                    _goldenMemManager->inPmem(diffInfo.physEffAddr)) {
+                    uint8_t current_golden[16] = {};
+                    _goldenMemManager->readGoldenMem(
+                        diffInfo.physEffAddr, current_golden, diffInfo.effSize);
+
+                    uint8_t gem5_pair[16];
+                    memcpy(gem5_pair,     &gem5_val, 8);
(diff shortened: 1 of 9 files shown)


XDu5 (changes requested):
Expanding issue_golden, goldenData, amoOldGoldenValue, and tmp_data, and replacing the 64-bit-only readGoldenMem() implementation are all necessary changes. However, the current patch removes the original panic but does not yet provide correct amocas.q behavior. The following issues should be fixed before merging:
1. The upper 64 bits of golden memory are not updated:
In LSQUnit::completeStore(), the AMO update still passes 0xff as the byte mask, updateGoldenMem() interprets each mask bit as one byte. Therefore, 0xff updates only bytes [0, 7]. For a successful 16-byte amocas.q, bytes [8, 15] remain stale.

2. The second destination is compared against the low half:
BaseCPU::diffWithNEMU() checks each destination register in a loop, but the AMO recovery comparison always starts at byte zero. The current implementation compares both destinations with old_value[0..7]. Please handle the 128-bit result as one pair. For example, construct the GEM5 result from both scalar destinations and compare all 16 bytes once. The memory and both destination registers should be synchronized only after the complete pair matches. This avoids partially repairing an amocas.q result.

3. readGoldenMem() does not validate the complete span:
The new implementation checks only the starting address:
if (inPmem(addr)) {
    std::memcpy(data, &goldenMem[addr - pmemBase], len);
}
A request can start inside PMEM but extend beyond its end. A guard-page test that reads 16 bytes starting eight bytes before the PMEM end terminates with SIGSEGV. Please validate the entire interval without overflow:
const bool valid =
    addr >= pmemBase &&
    addr - pmemBase <= pmemSize &&
    len <= pmemSize - (addr - pmemBase);

if (!valid) {
    warn(...);
    std::memset(data, 0, len);
    return;
}

std::memcpy(data, goldenMem + (addr - pmemBase), len);
It may be useful to provide an inPmem(addr, len) overload so all callers use the same range validation.

4. The CSR initialization change introduced by commit `4ca2c4bfa6` is only being used as a temporary prerequisite for local SPEC17 slice validation. It is not part of the final AMOCAS.Q fix. Please do not include this commit in the next revision of this PR. Before updating the PR, rebase the AMO-related changes onto the latest `xs-dev`/mainline and use the CSR initialization implementation from the updated mainline as the source of truth. It is fine to keep this commit in the current test branch so that the SPEC17 slice can be executed, but the final submitted patch should not carry this independent CSR change.

civil233:
Complete the 128-bit AMO (Zacas amocas.q) difftest path so that the
write -> read -> compare stages are consistent end to end.

Updated the code based on review feedback：

  1. lsq_unit.cc: fix golden_mask to update 64 bits of golden memory
    
  2. golden_global_mem: bounds-check readGoldenMem over [addr, addr+len)
      with an inPmem(addr, len) overload.
    
  3. base.cc (diffWithNEMU): pair-compare 128-bit AMO destination
      registers. At dest_idx==0, join scalarResults[0..1] into a 16-byte
      gem5_pair and memcmp against amoOldGoldenValue. A match means the
      GEM5-loaded old value was the globally-visible value at issue time,
      so the NEMU/GEM5 difference is just memory ordering: sync the
      post-AMO golden value and both registers to the reference and skip

XDu5 (approved):

jensen-yan (approved):
LGTM，就是DPRINTF 和调试输出稍微有点多，可以精简一点，或者就这样也行

PR #1108, opened 2026-09-02
cpu: Clarify difftest lifecycle and handle initial events

jensen-yan:
## Problem and behavior

This PR builds on #1106. The reference must start from the saved DUT pre-event state before processing its first architectural event, including an O3 trap or interrupt. Previously only the first comparable commit initialized REF, and a first MMIO skip could write an unpopulated host-side reference buffer back to NEMU.

The change preserves the dedicated startup snapshot and makes reference initialization an idempotent event-boundary operation. O3 initializes REF before handling an architectural fault; interrupt injection uses the same guard. Initial-PC validation uses the pre-event PC, so a trap handler does not overwrite or invalidate the startup snapshot. Initialization reads back NEMU's canonical state, and MMIO skip reads REF before updating the next PC and a nonzero scalar destination. SimpleCPU CSR capture now selects the requested thread.

## Readability and scope

- Move CPU difftest implementation from `base.cc` into `difftest_cpu.cc`; keep proxy/ABI handling in `difftest.cc` and `difftest.hh`.
- Separate REF stepping from result comparison. The existing PC/register/CSR/vector/shared-memory comparison body is preserved, apart from a shortened comment.
- Retain the earlier cleanup: separate `initialDutState`, explicit register-copy direction wrappers, and removal of unused proxy hooks/state.
- Document buffer ownership, synthesized CSR reads, initialization order, and current checking limitations in `docs/Gem5_Docs/top/difftest.md`.

This now includes correctness fixes and is no longer a behavior-neutral refactor. CSR comparison coverage and waiver policy, the compact NEMU ABI, and the existing non-guided exception retry policy remain unchanged. Extending those belongs in a follow-up PR.

## Validation

- `git diff --check` and repository style checks passed.
- Optimized build passed: `scons build/RISCV/gem5.opt --gold-linker -j64`.
- First-instruction increment smoke passed with both the pinned FS=0 REF and the FS=1 REF.
- Reset PC `0x80000002` smoke passed.
- First-event MMIO byte load: old head failed difftest on the next instruction; final build passed 10 commits.
- First-event instruction page fault: old head crashed inside REF before a normal commit; final build reached the handler and passed 10 commits.
- CoreMark with a 10K instruction limit passed with both REFs. Baseline and both final runs matched: 10,003 guest instructions, 62,245 CPU cycles, 20,727,252 ticks, and 15,755 integer-register reads.

Temporary startup fixtures use GDB to set initial GPR/CSR values before snapshot capture. The page-fault fixture also configures REF PMP permissions, which are outside the compact register snapshot. These fixtures are local validation, not new repository test infrastructure.

The pending-interrupt smoke takes the interrupt after one normal commit; it does not establish runtime coverage for an interrupt before every commit. Multi-context runtime coverage and the broader SimpleCPU exception/MMIO flow remain outside the validated scope.

## Stack

Base: #1106 (`codex/difftest-initial-state-sync`). Retarget to `xs-dev` after the base merges.

## Summary by CodeRabbit

- **New Features**
  - Improved RISC-V differential testing with broader register and control-state comparisons.
  - Added support for more reliable reference initialization during startup, traps, interrupts, and committed instructions.
  - Enhanced detection and handling of memory, atomic-operation, and cache-refill mismatches.

- **Bug Fixes**
  - Improved thread-specific register and control-register reads.
  - Added recovery and retry handling for certain instruction-state mismatches.

- **Documentation**
  - Reworked differential-testing documentation into a concise overview of behavior, boundaries, limitations, and verification priorities.


jensen-yan:
仅仅只是重构，简化下代码，没有功能性变化

PR #1116, opened 2026-09-03
cpu-o3: Align FP divide resources and timing with XiangShan RTL

XXtaoo:
# Motivation

The XiangShan RTL has two independent non-pipelined scalar FP divider instances in FEX1 and FEX2. The previous GEM5 configuration exposed only one FP_SLOW divider, which serialized independent fdiv.d instructions and caused large performance distortion on FDiv-intensive workloads.

The RTL divider also has operand-dependent completion paths, while GEM5 used a single fixed latency for all scalar FP divides.

## Changes

- Map FP_SLOW to fpIQ1 and fpIQ2, matching the two RTL divider units.
- Mark FloatDiv as non-pipelined so each divider remains occupied until result bypass/writeback.
- Model operand-dependent latency for 64-bit floating-point division across special-value, power-of-two, normal finite, and subnormal-related paths.
- Keep scheduling and completion timing consistent for each dynamic divide.

## Validation

Build:

    scons build/RISCV/gem5.opt --gold-linker -j64

40M-instruction checkpoint runs:

| Slice | Baseline GEM5 IPC | Updated GEM5 IPC | RTL IPC | Baseline vs RTL | Updated vs RTL |
|---|---:|---:|---:|---:|---:|
| leslie3d_62308 | 4.776399 | 3.139892 | 3.342665 | +42.89% | -6.07% |
| leslie3d_18494 | 4.472812 | 3.054794 | 3.144603 | +42.24% | -2.86% |

Baseline GEM5 IPC is calculated from the normal m5out/stats.txt results collected before this change. Each GEM5 IPC combines the two 20M-instruction statistics sections, so both GEM5 columns use the same 40M-instruction scope as the RTL results. This change reduces the GEM5-to-RTL gap from more than 42% above RTL to within 6.07% and 2.86% below RTL for the two slices.

## Scope

This is a behavior-level model of divider resource count and main operand-dependent completion paths. RTL outValidAhead3Cycle speculative wakeup is not enabled in GEM5 because the current model does not provide a corresponding early result-data bypass path.

## Summary by CodeRabbit

- **Performance**
  - Improved simulation accuracy for 64-bit floating-point division by accounting for operand-dependent execution latency.
  - Updated floating-point division scheduling to reflect non-pipelined divider behavior and result availability.
  - Consolidated floating-point execution resources to better represent how division operations are handled across scheduler ports.

### DIFF
--- a/configs/common/FUScheduler.py
+++ b/configs/common/FUScheduler.py
@@ -243,16 +243,14 @@ class KMHV3Scheduler(Scheduler):
     __fpIQs = [
         IssueQue(name='fpIQ0', inports=2, size=18, oports=[
             IssuePort(fu=[FP_ALU(), FP_MISC(), FP_MAC()],
-                      rp=[FpRD(0,0), FpRD(1, 0), FpRD(2,0)]),
-            IssuePort(fu=[FP_SLOW()],
-                      rp=[FpRD(2,1), FpRD(5,1)])
+                      rp=[FpRD(0,0), FpRD(1, 0), FpRD(2,0)])
         ]),
         IssueQue(name='fpIQ1', inports=2, size=18, oports=[
-            IssuePort(fu=[FP_ALU(), FP_MAC()],
+            IssuePort(fu=[FP_ALU(), FP_MAC(), FP_SLOW()],
                       rp=[FpRD(3,0), FpRD(4,0), FpRD(5,0)])
         ]),
         IssueQue(name='fpIQ2', inports=2, size=18, oports=[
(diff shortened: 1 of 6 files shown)


jueshiwenli:
Looks great. I’ll add some microbenchmarks to test it.

happy-lx (approved):

jueshiwenli (approved):
Image

jensen-yan:
合入该PR 后，浮点 kmhv3 0.3c 从22.9 降低到21.7分，降低1分多，性能影响显著

PR #1140, opened 2026-09-10
cpu-o3: Calibrated SBUFFER behaviour

jueshiwenli:
Thank  @ChrisYzcc  for identifying the mismatch.
Release SBuffer entries on accepted store misses.
Release the physical SBuffer entry when an L1D store miss is accepted by the MSHR, while retaining a separate snapshot for the in-flight request and store-to-load forwarding.

Track pending store misses for same-line ordering, forwarding, fence, flush, and drain handling. Preserve the legacy behavior through the sbufferReleaseOnMiss parameter.

Change-Id: Ia5ebf53a384dbcebebf36761facdcd8dca7daaa2

## Summary by CodeRabbit

- **New Features**
  - Store-buffer entries can now be released when an L1 data-cache miss is accepted, improving resource availability while preserving load forwarding and replay behavior.
  - Added a configurable option to control this behavior, enabled by default.
  - Added statistics for tracking store-buffer misses, pending entries, forwarding, and same-line replays.

- **Documentation**
  - Added detailed documentation covering store-buffer and MSHR lifetimes, forwarding, and fence, flush, and drain behavior.

### DIFF
--- a/src/cpu/o3/lsq.cc
+++ b/src/cpu/o3/lsq.cc
@@ -142,6 +142,10 @@ LSQ::StoreBufferEntry::recordForward(RequestPtr req, LSQRequest *lsqreq,
         assert(offset == 0);
     }
     bool full_forward = true;
+    auto *lsq = lsqreq->_port.getLsq();
+    auto *inflight_entry = lsq->find_inflight_store_buffer_entry(
+        blockPaddr, load_tid);
+    bool miss_forward = false;
     auto byteEligible = [&](StoreBufferEntry *entry, int byte_idx) {
         return entry && entry->tid == load_tid && entry->seqNum < load_seq &&
                entry->validMask[byte_idx];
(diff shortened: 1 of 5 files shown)


src/cpu/o3/lsq.cc
```diff
@@ -160,11 +164,21 @@ LSQ::StoreBufferEntry::recordForward(RequestPtr req, LSQRequest *lsqreq,
             lsqreq->SBforwardPackets.push_back(
                 LSQRequest::FWDPacket{
                     .idx = goffset + i, .byte = blockDatas[offset + i]});
+            miss_forward |= this == inflight_entry;
+        } else if (byteEligible(inflight_entry, offset + i)) {
+            lsqreq->SBforwardPackets.push_back(
+                LSQRequest::FWDPacket{
+                    .idx = goffset + i,
+                    .byte = inflight_entry->blockDatas[offset + i]});
+            miss_forward = true;
```
ChrisYzcc: It seems that the simulator does the MSHR store data forward in sbuffer, which is slightly different from the RTL behavior. Since L1D may block the query in RTL, this could make the simulation model less accurate.

I'm not an expert in GEM5, so this is just a suggestion. Feel free to resolve this comment if you think the current modeling is reasonable.

ChrisYzcc (commented):
Is the score for `lbm` within the expected range?

ChrisYzcc (approved):

jueshiwenli:
> Is the score for `lbm` within the expected range?

The specifics still need to be tested after the changes are merged, but this is indeed an area where the two implementations are not aligned.

PR #1152, opened 2026-09-15
misc: Reject guest kernel panics before performance scoring

jensen-yan:
A guest kernel panic can leave GEM5 executing the panic loop until it reaches the instruction limit and exits with status 0. In SPEC17 SMT run 34822992124, all 1914 workloads were marked completed even though 237 guest logs contained a kernel panic, contaminating the reported scores.

Add a pre-scoring workflow step that checks for the explicit `Kernel panic - not syncing:` message. Affected workloads receive an `abort` marker, lose their `completed` marker, and are listed in the Actions summary. The job fails before score generation; existing archive steps still preserve the logs.

The check is embedded in the reusable workflow so it works for historical tested SHAs that do not contain a new helper script, and covers both local and distributed execution. It scans finished logs; it does not terminate simulations early or classify generic warnings as panics.

Validation: three tests execute the actual workflow step against zero-exit panics, ordinary warnings/existing failures, and a missing archive directory. All passed. The detection was also checked against copies of the actual 1914 logs and identified all 237 known panics; the original archive was not modified.

## Summary by CodeRabbit

* **Bug Fixes**
  * Performance workflows now detect guest kernel panics and mark affected runs as aborted instead of completed.
  * Workflows now fail when expected performance results are missing.
  * Failure summaries include up to 10 affected workloads for easier diagnosis.

* **Tests**
  * Added coverage for kernel panic detection, result marker handling, repeated execution, and missing performance data.


jensen-yan:
方法有点丑陋，而且不是经常出现，先暂时关闭吧，后续有需要再检查

PR #1164, opened 2026-09-15
cpu-o3,mem-cache: Add hit-under-block and per-thread fetch gate

YihengChen1:
### Motivation

This PR investigates the performance impact of splitting the fetch-side `cacheBlocked` flag and `retryPkt` queue into per-thread structures under the 2BPU 2Fetch (`smtNumPredictingThreads=2`, `smtNumFetchTargetThreads=2`) SMT configuration. During the analysis, we discovered that the cache side (`BaseCache`) does not support serving clean read hits while the cache is in the `blocked` state (MSHR-exhausted / downstream back-pressure)—all requests, including hits, are NAK'd at `tryTiming()` before hit/miss determination. This "hit-under-blocked" gap suppresses cross-thread hit-under-miss opportunities. The PR adds two independent, param-gated changes to enable controlled experimentation.

---

### Change 1: Cache-side hit-under-block (HUB) switch

Adds a `hit_under_block` Bool param (default `False`) to `BaseCache` (`Cache.py`). When enabled, `CpuSidePort::tryTiming()` probes for a clean read hit (via the new `BaseCache::probeHit()`, a side-effect-free `tags->findBlock` + `ReadableBit` check) when the cache is `blocked`. If it's a clean read hit, the request is let through to `handleTimingReqHit()` (served without MSHR allocation, blocked state not worsened); misses still NAK + retry. A `hitUnderBlockServed` stat counts served hits. The gate is per-cache: enable on icache and/or dcache independently via config.

**Files**: `src/mem/cache/Cache.py`, `src/mem/cache/base.hh`, `src/mem/cache/base.cc`

---

### Change 2: Fetch per-thread cacheBlocked/retryPkt param-gate

Adds a `fetch_per_thread` Bool param (default `False`) to `BaseO3CPU` (`BaseO3CPU.py`). The `cacheBlocked` flag and `retryPkt` queue are now always arrays indexed by `fetchSlot(tid)` (`= tid` if per-thread, `= 0` if global). When `fetch_per_thread=true` (per-thread), each thread has its own flag/queue—thread A's NAK no longer gates thread B (enabling cross-thread hit-under-miss at the fetch level), and the round-robin retry drain eliminates cross-thread head-of-line (HoL) blocking. When `fetch_per_thread=false` (global, default), all threads share slot 0—identical to the original single-flag/single-queue behavior (the round-robin drain degenerates to single-FIFO). The `doSquash` and sibling-bypass paths use contextId-checked logic that is correct in both modes.

**Files**: `src/cpu/o3/BaseO3CPU.py`, `src/cpu/o3/fetch.hh`, `src/cpu/o3/fetch.cc`

---

### Usage

```python
# Config example (smt_idealkmhv3 variant):
cpu.icache.hit_under_block = True       # serve icache hits under blocked
cpu.dcache.hit_under_block = True      # serve dcache hits under blocked
cpu.fetch_per_thread = True             # per-thread cacheBlocked/retryPkt (false=global)
```

4 experiment configs provided: `smt_idealkmhv3_{clean,fixed,fixed_dcache,hubglobal}.py` setting (icache_hub, dcache_hub, fetch_per_thread) = (F,F,F) / (T,F,T) / (T,T,T) / (T,T,F).

| Config | icache HUB | dcache HUB | fetch_per_thread |
|--------|:----------:|:----------:|:----------------:|
| clean | F | F | F |
| fixed | T | F | T |
| fixed_dcache | T | T | T |
| hub_global | T | T | F |

---

### Test Results

Full SPEC06 SMT regression (1169 slices × 4 variants) on `dd9592f9d7`. All variants complete cleanly (0 DRAMsim3 errors, 0 assert).

**SPECint scores (higher = faster, base = pg2_clean):**

| Config | SPECint | vs clean | Overall | Completed |
|--------|---------|----------|---------|:---------:|
| clean (no hub, global fetch) | 26.426 | — | 24.792 | 1165/1169 |
| fixed (icache hub + per-thread) | 26.455 | +0.11% | 24.811 | 1166/1169 |
| fixed_dcache (icache+dcache hub + per-thread) | 26.455 | +0.11% | 24.811 | 1167/1169 |
| **hub_global (icache+dcache hub + global fetch)** | **26.478** | **+0.20%** | 24.795 | 1164/1169 |

**Key observations:**

1. **HUB alone improves SPECint**: `hub_global` (HUB + global fetch) achieves the highest SPECint (26.478, +0.20% vs clean), confirming cache-side hit-under-block recovers real throughput on icache-pressure workloads.

2. **Per-thread fetch degrades SPECint**: `fixed_dcache` (26.455) vs `hub_global` (26.478) → per-thread fetch costs ~0.09% SPECint. Despite enabling "fairer" thread scheduling, per-thread fetch is net-negative.

3. **HUB activation is massive**: icache ~4.5M + dcache ~13M hits served under blocked across the regression (80% of slices have dcache MSHR exhaustion).

4. **Overall scores within noise** (±0.1%): hit-under-block is latency-recovery (each hit saves ~1–2 cycles; the blocked root cause—miss exhaustion—is unchanged), so aggregate impact is small.

---

### Why Per-Thread Fetch Degrades Performance

Per-thread fetch lets thread B continue issuing fetch requests during thread A's icache stall (MSHR exhaustion). Intuitively this should be net-positive (B makes progress instead of stalling). However, the data shows it is net-negative, for two reasons:

1. **Shared dcache congestion**: When A's icache stalls, A's backend (in-flight loads/stores) continues hitting the shared dcache. B continuing to fetch → B's pipeline advances → B issues additional dcache requests on top of A's in-flight ones. The dcache is a shared single-MSHR-pool resource; for dcache-sensitive workloads (gcc_s04, milc, GemsFDTD), it is near saturation. B's added pressure causes MSHR exhaustion more frequently (dcache `blockedCycles` +8% on degraded slices), delaying both threads' dcache misses → both threads' IPC drop ~1.3%. The marginal fetch progress B gains is outweighed by the congestion penalty both threads suffer—the classic SMT "greedy scheduling on a saturated shared resource" anti-pattern.

2. **Latency-recovery vs capacity-recovery**: The cache-side HUB only recovers *hit latency* during blocked periods (a hit would have been served ~1–2 cycles later anyway, once an MSHR frees). It does not relieve the blocked root cause (miss exhaustion). So the HUB's benefit is bounded, and per-thread fetch's congestion cost (prolonging the blocked state) exceeds the HUB's recovery on dcache-sensitive workloads.

`hub_global` (HUB + global fetch) avoids both: when A's icache stalls, B pauses → the dcache has spare capacity (only A's in-flight requests) → A's critical miss completes faster → stall shorter → both threads benefit. This "cooperative stall" (pausing the non-stalled thread to let the bottleneck clear) outperforms per-thread "greedy" scheduling on saturated shared resources.

The above analysis is a reasonable hypothesis inferred from the current stats data (dcache `blockedCycles`, per-thread IPC deltas, `hitUnderBlockServed` counts). Since the overall performance impact is within noise (±0.1% geomean, ~0.09% SPECint), we did not conduct a deeper investigation into the exact mechanism (e.g., per-slice dcache MSHR occupancy traces, event-ordering analysis, or cycle-by-cycle interleaving). The hypothesis is consistent with the observed data but should be validated with more targeted experiments if the performance gap becomes significant in future configurations.

---

### Files Changed

| File | Change |
|------|--------|
| `src/mem/cache/Cache.py` | +`hit_under_block` param |
| `src/mem/cache/base.hh` | +`hitUnderBlock` member, `probeHit()`, `incHitUnderBlockServed()`, `hitUnderBlockServed` stat |
| `src/mem/cache/base.cc` | +ctor init, +`ADD_STAT`, +`probeHit` impl, `tryTiming` hit-first in blocked branch |
| `src/cpu/o3/BaseO3CPU.py` | +`fetch_per_thread` param |
| `src/cpu/o3/fetch.hh` | `cacheBlocked[MaxThreads]`, `retryPkt[MaxThreads]`, `fetchPerThread`, `fetchSlot()` accessor |
| `src/cpu/o3/fetch.cc` | ctor init, all sites accessor-ized, doSquash contextId loop, sibling-bypass sameThreadRetry, round-robin drain |
| `configs/example/smt_idealkmhv3.py` | +hub + fetch_per_thread settings in `setDualFrontendProbeParams` |

## Summary by CodeRabbit

- **New Features**
  - Added an optional per-thread fetch retry mode for SMT CPUs, reducing interference between threads during instruction-cache stalls while preserving shared behavior by default.
  - Added an optional cache mode that serves clean read hits even when the cache is blocked by resource or downstream pressure. Misses continue to wait and retry normally.
  - Added statistics for read hits served while the cache is blocked.


jensen-yan:
Thanks for investigating this and adding the experimental switches! We have confirmed that both the RTL ICache and DCache support hit-under-block, so adding this capability to the model makes sense. Once the implementation is fixed and validated, we can consider enabling HUB by default in the XiangShan configurations.

I would prefer splitting this into two PRs if practical: one for cache-side HUB support, and another for per-thread Fetch blocking/retry handling. That would let us validate correctness and measure the impact of each change separately. Keeping them together is also reasonable if there is a strong implementation dependency, but the following issues should be addressed first.

**1. HUB must exclude requests that still require miss resources despite a tag match.**

In [`BaseCache::probeHit()`](https://github.com/OpenXiangShan/GEM5/blob/eafc625ce2fad681b9395a675db0de6fe2e24621/src/mem/cache/base.cc#L655), the predicate checks `isRead()`, `needsWritable()`, and the block's `ReadableBit`, but does not exclude `pkt->req->isUncacheable()`.

An uncacheable read can match an existing readable cache line and therefore pass this probe. However, [`Cache::access()`](https://github.com/OpenXiangShan/GEM5/blob/eafc625ce2fad681b9395a675db0de6fe2e24621/src/mem/cache/cache.cc#L168) explicitly evicts an existing block for an uncacheable access and returns a miss. The [uncacheable-read miss path](https://github.com/OpenXiangShan/GEM5/blob/eafc625ce2fad681b9395a675db0de6fe2e24621/src/mem/cache/cache.cc#L337) then allocates a new MSHR.

The problematic sequence is:

```text
MSHRs exhausted, cache blocked
  -> uncacheable read matches an existing readable line
  -> probeHit() admits it through the blocked gate
  -> Cache::access() treats it as an uncacheable miss
  -> allocateMissBuffer() needs an unavailable MSHR
```

This breaks the assumption that every request admitted by HUB can complete without allocating an MSHR, and can reach the empty-free-list assertion in MSHR allocation. This is a source-level finding; I have not reproduced it in a full simulation.

Please at least reject uncacheable requests in the HUB predicate and add directed coverage for a normal cacheable read hit, a miss, and an uncacheable read to an address with a resident line while MSHRs are full. More generally, the probe should agree with the actual access path about whether the request can be handled without the exhausted resources.

**2. The retry drain must stop after the first rejected send, including in the default shared mode.**

In [`Fetch::retryPendingIcacheRequests()`](https://github.com/OpenXiangShan/GEM5/blob/eafc625ce2fad681b9395a675db0de6fe2e24621/src/cpu/o3/fetch.cc#L2974), `break` only exits the current queue's loop. After `sendTimingReq()` returns false, subsequent thread iterations can still send through the same `icachePort`. If an earlier send succeeded, `progress` also allows another outer pass before a new retry callback.

With `fetch_per_thread=false`, every `fetchSlot(t)` is zero, so the next thread iteration can immediately retry the very same rejected packet. Thus the default mode does not preserve the previous stop-on-failure behavior.

The [RequestPort contract](https://github.com/OpenXiangShan/GEM5/blob/eafc625ce2fad681b9395a675db0de6fe2e24621/src/mem/port.hh#L203) requires waiting for `recvReqRetry()` after a failed send. Please return from the drain on the first rejection and iterate the shared slot only once.

**3. Fixing the drain alone does not cover new requests from another thread.**

In [`Fetch::handleSuccessfulTranslation()`](https://github.com/OpenXiangShan/GEM5/blob/eafc625ce2fad681b9395a675db0de6fe2e24621/src/cpu/o3/fetch.cc#L1230), the gate now checks only `cacheBlocked[fetchSlot(tid)]`. In per-thread mode, if thread A's request is rejected, only A's flag is set. Thread B can then complete translation and send a new request through the same `icachePort` before any retry callback.

Per-thread pending queues are reasonable, but thread-local waiting state and the shared port's outstanding retry obligation need to be represented separately. Allowing hits through on the cache side does not by itself change the existing RequestPort contract. Please cover both retry-drain sends and new-request sends in a directed test that rejects one request and verifies that no further request is sent on that port before the next retry callback.

My preference would be to land the cache-side HUB support with the boundary conditions and tests addressed first, and handle per-thread Fetch scheduling in a follow-up PR.

happy-lx:
> An uncacheable read can match an existing readable cache line and therefore pass this probe. However, [`Cache::access()`](https://github.com/OpenXiangShan/GEM5/blob/eafc625ce2fad681b9395a675db0de6fe2e24621/src/mem/cache/cache.cc#L168) explicitly evicts an existing block for an uncacheable access and returns a miss. The [uncacheable-read miss path](https://github.com/OpenXiangShan/GEM5/blob/eafc625ce2fad681b9395a675db0de6fe2e24621/src/mem/cache/cache.cc#L337) then allocates a new MSHR.

An uncache request typically won't hit a cache tag. I'm not sure if this scenario actually exists, but just to be on the safe side, I think it's worth adding this condition.

> We have confirmed that both the RTL ICache and DCache support hit-under-block, so adding this capability to the model makes sense. Once the implementation is fixed and validated, we can consider enabling HUB by default in the XiangShan configurations.

This implementation looks fine, and I also recommend that HUB can be enabled by default to align with RTL behavior.

YihengChen1:
Thanks for the detailed review! Per your suggestion, we've split this into two PRs. This commit contains only the cache-side HUB support with the probeHit boundary fix. Per-thread Fetch scheduling (points 2 & 3) will be submitted as a follow-up PR.

---

## Point 1: probeHit boundary fix

Fixed. `probeHit()` now rejects three request types that `access()` would treat as miss despite a tag match:

```cpp
if (pkt->req->isUncacheable())       return false;  // access() evicts + miss + MSHR
if (pkt->req->isCacheMaintenance())  return false;  // always forwarded
if (pkt->isEviction())               return false;  // writeback path
```

The probe now agrees with `access()` on whether a request can complete without an MSHR.

## Points 2 & 3: Deferred to follow-up PR

The per-thread Fetch changes (`fetch.cc`, `fetch.hh`, `BaseO3CPU.py`) have been removed from this PR. The original global `cacheBlocked` / single-FIFO `retryPendingIcacheRequests` (with `return` on first rejection) is unchanged — no protocol violation. These will be addressed in a separate PR with the shared `portRetryPending` flag you suggested.

---

## Regression Results

Full SPEC06 SMT regression (1169 slices) on `dd9592f9d7`. 1164 completed, 3 abort (pre-existing `setNeedReplay()` LSQ bug,  2 are actually nonexistent，confirmed reproducible on clean baseline without HUB — unrelated).

**HUB improvement vs baseline:**

| | SPECint | vs clean |
|---|---|---|
| clean (no HUB) | 26.43 | — |
| **this PR (HUB + probeHit fix)** | **26.48** | **+0.20%** |

Overall within noise (±0.1%). Hit-under-block is latency-recovery — each hit saves ~1-2 cycles; the blocked root cause (miss exhaustion) is unchanged.

## Stats evidence for correct operation

**Scenario 1 — cacheable read hit served under blocked:**
- icache `hitUnderBlockServed`: 11.6M hits
- dcache `hitUnderBlockServed`: 26.7M hits
- 1166/1169 slices have icache HUB activation, 1064/1169 have dcache activation

**Scenario 2 — cacheable read miss correctly rejected (NAK + retry):**
- icache `blockedCauses::no_mshrs`: 3.25M MSHR exhaustion events
- dcache `blockedCauses::no_mshrs`: 45.1M MSHR exhaustion events
- `assert(!freeList.empty())` crashes: **0** — if any miss had been incorrectly admitted through HUB, `access()` would have tried `allocateMissBuffer()` on an exhausted pool and crashed. Zero crashes across 48M+ exhaustion events confirms all misses were correctly rejected.

**Scenario 3 — uncacheable read excluded by probeHit:**
- dcache `overallMshrUncacheable`: 305 uncacheable accesses observed in SPEC06 (e.g., astar_biglakes)
- `freeList` / uncacheable-related assertion failures: **0**
- The `isUncacheable()` check in `probeHit()` rejects uncacheable reads at the blocked gate, preventing them from reaching `access()` where they would be treated as miss and attempt MSHR allocation. No uncacheable-related failure occurred in any of the 1164 completed slices.

happy-lx (approved):

jensen-yan (approved):

PR #1025, opened 2026-08-06
feat(smt): add per-IQ Watermark partitioning policy for SMT

TFS-star:
## Motivation

Under SMT, the default Dynamic IQ dispatch policy allows one thread to monopolize all IQ entries, starving the other thread. This is especially problematic for memory-intensive workloads where cache-missed instructions accumulate in the IQ waiting for data return, blocking the other thread from dispatching.

## Approach

Introduce a **Watermark** policy for SMT IQ dispatch that reserves a minimum number of entries per hardware thread. Each IQ instance carries a per-thread watermark value configured through CLI arguments (`--smt-iq-policy`, `--smt-iq-watermark-int/mem/fp`). The Python config layer classifies IQs by name prefix (int/fp/vec/mem) and binds the appropriate watermark, so the C++ side needs no string-matching logic.

The `ready(tid)` admission check computes a per-thread limit based on the other threads' actual occupancy and the watermark floor:

```
wmLimit = iqsize - sum(max(watermark, entries[t]) for t != tid)
```

- `watermark=0` (default) preserves the existing Dynamic behavior
- Negative watermark values are clamped to 0 with a warning to prevent uint32 overflow deadlock
- Unsupported policies (Partitioned/Threshold/DynamicBorrowing) fall back to Dynamic with a warning

## Environment

- **Branch**: `xs-dev`
- **Base commit**: `6d3120361e` (mem: fix bug of vsq (#1019))
- **Config**: SMT-2, KMHV3 scheduler
- **Test**: SPEC06 INT, 12 workloads, 682 checkpoints

## Results

| Config | score/GHz | Improvement |
|--------|-----------|-------------|
| Dynamic (baseline) | 23.799 | - |
| Watermark int=2 mem=4 | **23.933** | **+0.56%** |

Per-workload highlights:
- **mcf**: +2.44%
- **omnetpp**: +1.87%
- **gcc**: +1.84%

Memory-sensitive workloads benefit most, consistent with the design goal of memIQ watermark suppressing load accumulation and preventing sibling thread starvation.

## Changes

6 files changed, +112 -17:
- `src/cpu/o3/FuncScheduler.py` — add `smtIQWatermark` SimObject parameter
- `src/cpu/o3/BaseO3CPU.py` — add `Watermark` enum value, change `smtIQPolicy` default to `Dynamic`
- `src/cpu/o3/issue_queue.hh` — add SMT IQ partitioning members, update `ready()` signature
[...]

PR #1175, opened 2026-09-18
feat(smt): add safe per-IQ watermark policy

TFS-star:
## Motivation

Under SMT, issue queues are shared by multiple hardware threads. With the default Dynamic policy, one thread may occupy most or all available IQ entries, preventing the sibling thread from dispatching new instructions.

This is particularly problematic for memory-intensive workloads: instructions waiting on cache misses may remain in the memory IQ for a long time, consuming entries and causing severe sibling-thread starvation.

A per-thread minimum reservation is therefore needed while preserving dynamic sharing of unused entries.

## Approach

Introduce a `Watermark` policy for SMT IQ admission.

Each issue queue has a configurable per-thread watermark. Before admitting an instruction from thread `tid`, `ready(tid)` reserves capacity for every other thread according to the larger of that thread's current IQ occupancy and the configured per-thread watermark:

```text
reserved = sum(max(watermark, entries[t])) for all t != tid
wmLimit  = max(iqsize - reserved, 0)
allow    = entries[tid] < wmLimit
```

This provides the following behavior:

- A lightly occupied thread retains at least `watermark` entries.
- Unused capacity remains dynamically shareable.
- A heavily occupied thread is restricted as sibling occupancy grows.
- `watermark=0` preserves Dynamic behavior.

Watermarks are configured per IQ through:

```text
--smt-iq-policy
--smt-iq-watermark-int
--smt-iq-watermark-mem
--smt-iq-watermark-fp
```

The Python configuration classifies IQs by name:

- `int*` uses the integer watermark.
- `fp*` and `vec*` use the floating-point/vector watermark.
- Load, store-address, store-data, and other memory IQs use the memory watermark.

This keeps IQ classification in the configuration layer and avoids name-matching logic in C++.

## Safety and Compatibility

This revision adds safeguards that were missing from PR #1025.

During IQ initialization, the configuration is checked using:

```text
numThreads * watermark <= iqsize
```

If the requested reservation exceeds the IQ capacity, the affected IQ prints a warning and explicitly falls back to Dynamic sharing. Equality remains valid; for example, two threads with watermark 8 in a 16-entry IQ form a valid equal partition.

This prevents an oversized watermark or increased thread count from making the admission limit zero for every thread and causing a dispatch deadlock.

Additional safeguards include:

- Negative watermarks fall back to Dynamic sharing.
- Unsupported IQ policies fall back to Dynamic sharing.
- Reservation calculations use `uint64_t` to prevent integer overflow.
- `IssueQue::ready()` and `Scheduler::ready()` consistently use `ThreadID`.

The last item fixes the compilation failure in PR #1025, where the header used `int` while the implementation used `ThreadID` (`int16_t`).

## Previous Performance Results

The following results were obtained by the earlier Watermark experiment on SPEC06 INT:

| Configuration | score/GHz | Improvement |
|---|---:|---:|
| Dynamic baseline | 23.799 | — |
| Watermark `int=2, mem=4` | 23.933 | +0.56% |

Per-workload highlights:

- `mcf`: +2.44%
- `omnetpp`: +1.87%
- `gcc`: +1.84%

Memory-sensitive workloads benefit most, consistent with the intended behavior: the memory-IQ watermark limits long-lived instruction accumulation and preserves dispatch capacity for the sibling thread.


jensen-yan:
当前配置默认没有开启IQ watermark，还是保留为Dynamic

jensen-yan (approved):

